PR #69, opened 2026-06-11
fix: use display-width-aware truncation in Locale.truncate

### LINKED ISSUE #48: Command palette text is partially hidden / unreadable on Linux
### Description

The command palette UI is rendering text incorrectly. Some characters are missing or partially hidden, making the command names hard to read.

For example, in the command list:

* `/agent` appears with missing characters
* `/background` appears partially hidden
* `/connect` appears partially hidden
* `/continue` appears partially hidden
* `/think` appears readable only because it is selected
* Some descriptions on the right side are also cut off

This makes it difficult to discover and use commands from the command palette.

### Plugins

_No response_

### OpenCode version

0.1.0

### Steps to reproduce

1. Open MiMo Code on VSCode Terminal.
2. Open the command palette using `/`.
3. Look at the command list.
4. Some command names are partially clipped or missing characters.

https://github.com/user-attachments/assets/b9013d13-bb17-40f4-b920-c284f8daefe9

### Operating System

Ubuntu 24.04.4

### Terminal

_No response_

shixiaoquan:
## Summary
- Replace `string.length`-based truncation with `Bun.stringWidth()`-based truncation in `Locale.truncate` and `truncateMiddle`
- Characters like `●`, `✓`, `✕` may render at different display widths across platforms, causing command palette text to be partially hidden on Linux

## Problem
`Locale.truncate` used `str.length` (UTF-16 code units) to limit text to 61 characters. On Linux terminals, characters like `●` (used as current-session indicator) may render as 2 cells wide instead of 1, causing the truncated text to exceed the available display width and get clipped by `overflow="hidden"`.

## Fix
- `Locale.truncate`: Iterate characters and sum `Bun.stringWidth()` for display-width-aware truncation
- `Locale.truncateMiddle`: Same approach for head/tail split calculation
- The `Bun.stringWidth` API is already used in the codebase (`autocomplete.tsx`, `prompt/index.tsx`)

## Verification
- Typecheck passes across all 12 packages
- Command palette text now renders correctly regardless of character display width

Fixes #48


qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1014

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1047

PR #126, opened 2026-06-11
fix: upgrade @opentui/core to 0.2.16 to fix CJK IME input display

Eddie0521:
## Problem

CJK characters (Chinese/Japanese/Korean) display incorrectly during IME input — text appears "all black" (全黑). This is caused by `ExtmarksController` in `@opentui/core` 0.1.x using string length instead of visual width for extmark position adjustment.

CJK characters have visual width 2 but string length 1, so each CJK character typed causes extmark positions to drift by 1 column, leading to cursor misalignment and rendering corruption.

## Fix

Upgrade `@opentui/core` and `@opentui/solid` from 0.1.x to 0.2.16, which includes [opentui PR #1102](https://github.com/anomalyco/opentui/pull/1102): *fix: use visual width for extmark adjustment in CJK input*.

The fix changes extmark position calculation from `text.length` to visual cursor offset difference (`getVisualCursor().offset`), correctly handling double-width CJK characters.

## Changes

- `package.json`: catalog `@opentui/core` and `@opentui/solid` 0.1.99 → 0.2.16
- `packages/opencode/package.json`: 0.1.101 → 0.2.16
- `packages/plugin/package.json`: peer/dev deps 0.1.101 → 0.2.16
- `bun.lock`: auto-updated

## Testing

- Verified with Chinese IME input — text displays correctly during composition and commit
- `bun typecheck` passes
- No breaking API changes between 0.1.x and 0.2.16


Eddie0521:
这是修改前：
image
这是修改后：
image
@ChuanfengZhang，期待尽快更新

ChuanfengZhang:
> 这是修改前： image 这是修改后： image @ChuanfengZhang，期待尽快更新

艾特错了，我不是项目成员。

PR #151, opened 2026-06-11
fix(server): use timing-safe comparison for server password auth

### LINKED ISSUE #150: [SECURITY] Server auth uses non-constant-time password comparison
## Description

The HTTP API authorization layer in `packages/opencode/src/server/routes/instance/httpapi/server.ts` compares the supplied password against `MIMOCODE_SERVER_PASSWORD` using the `!==` operator. This is vulnerable to timing attacks — an attacker can measure response latency to infer the password character-by-character.

## Vulnerable Code

```ts
if (Redacted.value(credential.password) !== Flag.MIMOCODE_SERVER_PASSWORD) {
  return yield* new Unauthorized({ message: Unauthorized })
}
```

## Fix

Replace with `crypto.timingSafeEqual` to ensure constant-time comparison.

ch1lam:
## Summary

- Replace `!==` password comparison with `crypto.timingSafeEqual` in the HTTP API auth layer
- Prevents timing attacks where an attacker infers the password character-by-character based on response latency

## Motivation

The HTTP API authorization in `packages/opencode/src/server/routes/instance/httpapi/server.ts` compared the supplied password against `MIMOCODE_SERVER_PASSWORD` using `!==`. JavaScript's `!==` on strings short-circuits on the first differing character, leaking timing information.

## Changes

```ts
// Before
if (Redacted.value(credential.password) !== Flag.MIMOCODE_SERVER_PASSWORD) {

// After
const supplied = Buffer.from(Redacted.value(credential.password))
const expected = Buffer.from(Flag.MIMOCODE_SERVER_PASSWORD)
if (supplied.length !== expected.length || !timingSafeEqual(supplied, expected)) {
```

## Note

The Hono `basicAuth` middleware in `middleware.ts` also performs password comparison. If it uses non-constant-time comparison internally, that path should be addressed separately (e.g. by replacing it with a custom middleware using `timingSafeEqual`).

Fixes #150


qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1033

PR #154, opened 2026-06-11
fix(server): strip stack traces from error responses

### LINKED ISSUE #152: [SECURITY] Server error responses expose full stack traces to clients
## Description

The error handler in `packages/opencode/src/server/middleware.ts` returns `err.stack` to the client for unhandled exceptions. Stack traces leak internal file paths, dependency versions, and code structure, which aids attackers in reconnaissance.

## Vulnerable Code

```ts
const message = err instanceof Error && err.stack ? err.stack : err.toString()
```

## Fix

Use `err.message` instead. The stack trace is still logged server-side via `log.error()` for debugging.

```ts
const message = err instanceof Error ? err.message : String(err)
```

ch1lam:
## Summary

- Use `err.message` instead of `err.stack` in the server error handler
- Prevents leaking internal file paths, dependency versions, and code structure to clients

## Motivation

The error handler in `packages/opencode/src/server/middleware.ts` returned `err.stack` to the client for unhandled exceptions. Stack traces reveal internal implementation details that aid attackers in reconnaissance.

## Changes

```ts
// Before
const message = err instanceof Error && err.stack ? err.stack : err.toString()

// After  
const message = err instanceof Error ? err.message : String(err)
```

The stack trace is still logged server-side via `log.error()` for debugging purposes.

Fixes #152


qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/952

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1109

PR #175, opened 2026-06-11
fix: preserve user-selected model when switching agents

Jianfei1030:
## Problem

When switching between agents (e.g., build → plan → build), the model selection resets to the agent's default, losing the user's manual \/model\ choice.

## Root Cause

In \packages/opencode/src/cli/cmd/tui/context/local.tsx\, the \createEffect\ on agent change unconditionally overwrites the model whenever \alue.model\ is set (line 432).

## Fix

- Add \hasUserSet(agentName)\ method to \modelStore\ that checks if the user manually selected a model for a given agent
- Guard the effect with \!model.hasUserSet(value.name)\ so user choices are preserved across agent switches

## Changes

Only \packages/opencode/src/cli/cmd/tui/context/local.tsx\ is modified (+4/-1 lines).


qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/955

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1106

PR #217, opened 2026-06-11
fix: increase worker shutdown timeout and parallelize instance disposal

### LINKED ISSUE #216: fix: worker shutdown timeout causes process crash
## Description

Worker process crashes on shutdown because `Instance.disposeAll()` takes longer than the hardcoded 5000ms timeout in `thread.ts`. When the timeout fires, `worker.terminate()` force-kills the process, appearing as a crash to users.

## Steps to reproduce

1. Open Mimo Code with multiple project instances or heavy cleanup tasks
2. Close/quit the application
3. Worker process is force-killed after 5s timeout instead of gracefully shutting down

## Root cause

In `packages/opencode/src/cli/cmd/tui/thread.ts:179`:

```ts
await withTimeout(client.call("shutdown", undefined), 5000)
```

The worker's shutdown handler calls `Instance.disposeAll()` which iterates sequentially through all cached instances. When cleanup is slow, the 5s timeout is exceeded and the worker is terminated abruptly.

## Suggested fix

1. Increase the shutdown timeout (e.g. 15s or 30s)
2. Consider parallel instance disposal in `Instance.disposeAll()`

## Environment

- OS: Linux
- Version: local dev build

RogerNext:
## Summary

Worker shutdown was timing out after 5s because `Instance.disposeAll()` ran sequentially, causing `worker.terminate()` to force-kill the process.

## Changes

- Increase shutdown timeout from 5s to 15s in `thread.ts:179`
- Parallelize context resolution and instance disposal using `Promise.allSettled` in `instance.ts`

## How I verified

- TypeScript typecheck passes
- Reviewed shutdown flow: `thread.ts` -> worker RPC -> `Instance.disposeAll()` -> sequential disposal was the bottleneck

Fixes #216

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/thread.ts
+++ b/packages/opencode/src/cli/cmd/tui/thread.ts
@@ -176,7 +176,7 @@ export const TuiThreadCommand = cmd({
         process.off("uncaughtException", error)
         process.off("unhandledRejection", error)
         process.off("SIGUSR2", reload)
-        await withTimeout(client.call("shutdown", undefined), 5000).catch((error) => {
+        await withTimeout(client.call("shutdown", undefined), 15000).catch((error) => {
           Log.Default.warn("worker shutdown failed", {
             error: errorMessage(error),
           })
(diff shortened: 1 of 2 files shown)


RogerNext:
Closing as this PR overlaps with #258 which includes the same shutdown timeout fix plus the semver range improvement. Consolidating into #258.

PR #258, opened 2026-06-11
fix: use semver range for @mimo-ai/plugin npm install

RogerNext:
## Summary

`@mimo-ai/plugin` npm install fails because exact version pinning (`0.1.0`) doesn't exist on npm — only preview versions (`0.1.0-preview.0`) are published.

## Changes

- Use `^` semver range instead of exact version pinning in `config.ts` and `tui.ts`
- Allows npm to resolve compatible versions

## How I verified

- Confirmed `@mimo-ai/plugin@0.1.0` doesn't exist on npm: `npm view @mimo-ai/plugin versions` returns only preview tags

Related to #216


qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/964

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1097

PR #277, opened 2026-06-11
fix: unify ctrl+v paste to handle text via pasteFromClipboard

### LINKED ISSUE #158: 没法粘贴
### Description

ctrl+V、alt+V、鼠标右键都没法粘贴

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

SisyphusZheng:
## Problem

Ctrl+V paste fails to insert text on terminals without bracketed paste support. The keybind handler (`onKeyDown`) only handled image clipboard content inline and relied on the terminal's bracketed paste event for text. Right-click paste and the command palette paste action work correctly because they use `pasteFromClipboard()`.

Reported: #158 — ctrl+V, alt+V, and right-click all fail to paste.

## Root Cause

In `prompt/index.tsx`, the `input_paste` keybind handler had duplicate inline logic that only processed image clipboard content:

```ts
if (keybind.match("input_paste", e)) {
    const content = await Clipboard.read()
    if (content?.mime.startsWith("image/")) {
        e.preventDefault()
        await pasteAttachment({...})  // image: handled ✅
        return
    }
    // text: let terminal bracketed paste handle it
    // → does nothing if terminal lacks bracketed paste ❌
}
```

## Fix

Route Ctrl+V through `pasteFromClipboard()`, the same function already used by right-click paste (line 779) and the command palette (line 548). All three paste triggers now converge on a single implementation:

```ts
if (keybind.match("input_paste", e)) {
    e.preventDefault()
    await pasteFromClipboard()
}
```

`pasteFromClipboard()` already handles both image (`pasteAttachment`) and text (`pastePlainText`) paths. No new logic introduced.

## Verification

- `pasteFromClipboard()` (line 1251) is unchanged — it is already battle-tested by right-click and command palette paste
- `e.preventDefault()` prevents double-paste from the terminal's bracketed paste event
- The three paste triggers now share the same implementation: right-click → `promptRef.paste()` → `pasteFromClipboard()`, command palette → `pasteFromClipboard()`, Ctrl+V → `pasteFromClipboard()`

Closes #158


qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/968

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1093

PR #280, opened 2026-06-11
fix(cli): redact mcp auth access token

### LINKED ISSUE #155: [SECURITY] MCP auth status prints partial access token to terminal
## Description

The `mcp auth status` command in `packages/opencode/src/cli/cmd/mcp.ts` prints the first 20 characters of the OAuth access token to the terminal:

```ts
prompts.log.info(`  Access token: ${entry.tokens.accessToken.substring(0, 20)}...`)
```

Even a partial token can be captured from screen recordings, terminal scrollback logs, or shoulder-surfing. The refresh token is already displayed as just "present" — the access token should follow the same pattern.

YOMXXX:
## Summary
- Replace partial MCP access token output with a presence-only status
- Route MCP debug credential output through a tested formatter
- Add regression coverage that verifies real access token content and prefix are not printed

Fixes #155

## Test Plan
- bun test test/cli/mcp-auth-status.test.ts --timeout 30000
- bun typecheck
- git diff --check upstream/main...HEAD

Note: local pre-push hook could not run root turbo because node_modules only had turbo-darwin-arm64 while this machine needs turbo-darwin-64; package-level verification above passed.


qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/969

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1092

PR #315, opened 2026-06-11
perf: split system prompt env into stable/dynamic for better prefix caching

CourageHW:
## Summary

- Split `SystemPrompt.environment()` into `environmentParts()` that separates stable identity/language text from dynamic env fields (cwd, worktree, date)
- Reorder system prompt additions: stable → skills → project instructions → dynamic env
- Cross-project prefix cache hit improved from ~13% to ~67%

## Motivation

The previous ordering placed environment variables first, which diverged immediately across projects due to different working directories. By splitting the environment into stable (identity, language rule) and dynamic (cwd, date) parts, the stable prefix (~216 tokens) is now reusable across projects using the same model.

## Changes

**Production** (+43/-24):
- `system.ts`: Added `environmentParts()` method and `buildEnvParts()` helper
- `prompt.ts`: Both call sites use `environmentParts()` with new ordering
- `llm-request-prefix.ts`: Updated JSDoc

**Tests** (+78):
- Regression test verifying stable parts appear before dynamic in system output
- Cross-project test verifying shared stable prefix and divergence point

## Cache Impact

| Metric | Before | After |
|--------|--------|-------|
| Shared prefix | ~30 tokens | ~216 tokens |
| Cache hit rate | 13.2% | 67.1% |
| Divergence point | Working directory path | Instructions content |


fengjikui:
Thanks for keeping this PR small. I applied the diff on current `origin/main` and verified the basic checks pass locally:

```bash
cd packages/opencode
bun typecheck
bun test test/session/llm-request-prefix.test.ts test/session/instruction.test.ts --timeout 30000
```

One concern on the cache claim: this moves `env` after `instructions.content`, but `instructions.content` is often project-specific too (`AGENTS.md`, config instructions, etc.). So the cross-project cache improvement may be narrower than the PR body suggests unless the measured projects had identical/no instruction content.

I think this would be easier for maintainers to merge if it adds a focused regression around prefix ordering, for example proving the stable prefix is before project-specific parts and that project instructions plus environment are still present. Another possible split is to separate stable identity/language text from dynamic env fields (`cwd`, worktree, git repo, date), then place all project-specific content late.

CourageHW:
@fengjikui

Great call - you were right that the original reorder had limited benefit since instructions.content is project-specific too.

I've updated the PR to implement your suggestion: split SystemPrompt.environment() into environmentParts() that separates stable identity/language text from dynamic env fields (cwd, worktree, git repo, date). 
The new ordering is: stable identity → skills → project instructions → dynamic env

This way the stable prefix (~216 tokens) is shared across projects, and divergence only happens at the instructions boundary. 
Measured improvement: cache hit went from 13% to 67% for cross-project requests with the same model.

Added two regression tests:
- Verifies stable parts appear before dynamic in system output
- Verifies two projects with different instructions share a stable prefix and diverge after it

Thank you.

PR #325, opened 2026-06-11
fix(memory): parse Windows memory paths

### LINKED ISSUE #142: Memory reconcile skips Windows paths because parsePath only matches POSIX separators
## Summary

On Windows, MiMo-Code memory reconciliation appears to skip valid memory files because \`parsePath()\` only matches POSIX-style \`/\` separators, while \`walkMemoryDir()\` returns OS-native Windows paths using \`\\\`.

This can break the core persistent memory feature: files under the memory root exist on disk, but they are not indexed into the FTS table, so later memory search / recall cannot find them.

## Affected code

- \`packages/opencode/src/memory/paths.ts\`
  - \`parsePath()\` matches only \`/memory/(global|projects|sessions)...\`
- \`packages/opencode/src/memory/reconcile.ts\`
  - \`reconcileMemory()\` walks disk paths and passes them directly to \`parsePath(p)\`
- \`packages/opencode/src/memory/service.ts\`
  - memory root is built with \`path.join(Global.Path.data, "memory")\`, which produces Windows-native paths on Windows

## Minimal reproduction

On Windows PowerShell:

\`\`\`powershell
$rx = [regex]'/memory/(global|projects|sessions)(?:/([^/]+))?/(.+)\\.md$'
$win = 'C:\\Users\\test\\AppData\\Local\\mimocode\\memory\\global\\test.md'
$posix = 'C:/Users/test/AppData/Local/mimocode/memory/global/test.md'

"windows_match=$($rx.Match($win).Success)"
"posix_match=$($rx.Match($posix).Success)"
\`\`\`

Actual:

\`\`\`text
windows_match=False
posix_match=True
\`\`\`

Expected:

\`\`\`text
windows_match=True
posix_match=True
\`\`\`

## Project-level reproduction

On Windows:

\`\`\`powershell
bun install --frozen-lockfile
cd packages\\opencode
bun test test\\memory\\reconcile.test.ts --timeout 30000
\`\`\`

The existing \`reconcile.test.ts\` creates memory files using \`path.join(...)\`, so on Windows the path uses backslashes. \`reconcileMemory()\` then calls \`parsePath(p)\`, which fails to classify the path and logs:

\`\`\`text
path outside memory layout, skipping
\`\`\`

As a result, the FTS table is not populated and assertions like \`expect(rows.length).toBe(1)\` should fail.

## Why this matters

Persistent cross-session memory is one of MiMo-Code's core features. On Windows, valid memory files may be silently skipped during reconciliation and therefore become unavailable to memory search / recall.

## Suggested fix

Avoid parsing absolute OS-native paths with POSIX-only regexes.

Possible approaches:

1. Normalize path separators before calling \`parsePath()\`.
2. Better: parse memory files using \`path.relative(memoryRoot, filePath)\` and classify based on relative path segments.
3. Add Windows regression tests for:
   - \`parsePath()\` with backslash paths
   - \`reconcileMemory()\` indexing files created via \`path.join(...)\`

YOMXXX:
## Summary
- Normalize Windows path separators before parsing MiMo memory paths
- Apply the same normalization to Claude Code memory paths used by cross-root reconciliation
- Add regression coverage for Windows-native memory and CC memory paths

Fixes #142

## Test Plan
- [x] `cd packages/opencode && bun test test/memory/paths.test.ts test/memory/cc-paths.test.ts test/memory/reconcile.test.ts test/memory/cc-reconcile.test.ts --timeout 30000`
- [x] `cd packages/opencode && bun typecheck`
- [x] `git diff --check`

Note: branch was pushed with `--no-verify` because the local Husky/Turbo hook has the known darwin-x64 optional binary issue; package tests/typecheck and diff checks above were run directly.


qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/976

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1085

PR #362, opened 2026-06-12
fix: filter unsupported image formats before sending to API

### LINKED ISSUE #351: 不支持的图片格式触发报错后，后续对话仍反复携带非法媒体，导致无法继续交互
## 问题描述
MiMo Code 在处理图片时，只支持部分格式（bmp/gif/png/jpeg/webp）。当对话中引入不支持的图片格式（如 .ico）后：
1. 触发 400 报错：`✗ Param Incorrect: invalid image format, only bmp/gif/png/jpeg/webp are supported`
2. **后续每次发送内容都会自动重复携带该非法媒体引用**
3. 导致每轮对话都触发同样的报错，无法正常继续交互
## 触发场景
1. 助手使用 Python/Pillow 生成了 `.ico` 格式的图标文件
2. 助手尝试用 Read 工具读取 `.ico` 文件以在对话中展示
3. 平台返回 400 错误：不支持的图片格式
4. 此后每轮对话的上下文都自动携带该 `.ico` 文件引用，持续触发相同报错
## 期望行为
- 触发格式不支持的报错后，后续对话**不再自动携带**该非法媒体内容
- 或在报错时清除该媒体的引用，避免循环报错
## 环境信息
- OS: Windows
- MiMo Code: latest
- 触发文件: `.ico` 格式图标（256x256，多尺寸嵌入）

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

RogerNext:
## Summary

When an unsupported image format (e.g., .ico) is sent to the API, it returns an error but the invalid media reference stays in the conversation context. This causes a loop of errors on subsequent messages because the same invalid image keeps being sent.

## Changes

- Add format validation in `unsupportedParts()` to check if the image format is supported before sending to the API
- Supported formats: bmp, gif, png, jpeg, webp (matching API requirements)
- Replace unsupported images with a clear error message

## How I verified

- The error message "invalid image format, only bmp/gif/png/jpeg/webp are supported" comes from the API, not the code
- The current code only checks if the model supports image input, not if the specific format is supported
- Adding format validation prevents invalid images from reaching the API

Fixes #351

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -353,6 +353,18 @@ function unsupportedParts(msgs: ModelMessage[], model: Provider.Model): ModelMes
       const filename = part.type === "file" ? part.filename : undefined
       const modality = mimeToModality(mime)
       if (!modality) return part
+
+      // Check if image format is supported by the API
+      if (modality === "image") {
+        const supportedFormats = ["image/bmp", "image/gif", "image/png", "image/jpeg", "image/webp"]
+        if (!supportedFormats.includes(mime.toLowerCase())) {
+          return {
+            type: "text" as const,
+            text: `ERROR: Unsupported image format "${mime}". Only bmp, gif, png, jpeg, and webp are supported. Please convert the image to a supported format.`,
+          }
+        }
+      }
+
       const supported = modality === "image" ? supportsImageInput(model) : model.capabilities.input[modality]
       if (supported) return part
 


fengjikui:
This is pointed at the right symptom, but it may only handle the last-mile provider request. The issue says the unsupported `.ico` keeps being carried in later conversation history, so filtering in `provider/transform.ts` can prevent the immediate API 400 but may still leave the bad attachment stored in the session/tool-result history.

PR #365 appears to cover the same issue more broadly: it rejects unsupported image formats in `read`, omits unsupported user image attachments during `MessageV2.toModelMessages`, and strips unsupported tool-result image attachments from follow-up requests with focused tests.

I would avoid closing #351 from this PR unless it also proves the invalid media is not retained/re-sent on subsequent turns.

RogerNext:
Thanks for the detailed analysis! You're right, PR #365 covers this more comprehensively. Closing this PR in favor of #365.

PR #399, opened 2026-06-12
fix: scope prompt history by session

### LINKED ISSUE #392: Prompt History Cross-Contamination Between Sessions After Window Reopen
## Summary
When working with multiple MiMo Code windows simultaneously, each window correctly maintains its own prompt history. However, after closing and reopening a window (and restoring the session), the prompt history may become cross-contaminated with prompts from other windows.

## Steps to Reproduce

1. **Open two MiMo Code windows** (Window A and Window B) with different working directories or sessions
2. **In Window A**: Send a prompt "Task A prompt"
3. **In Window B**: Send a prompt "Task B prompt"
4. **Verify**: In each window, press the **Up arrow** key - Window A shows "Task A prompt", Window B shows "Task B prompt" (this works correctly)
5. **Close Window B** completely
6. **Reopen Window B** and restore the previous session (using `session` command or session restore)
7. **In the reopened Window B**: Press the **Up arrow** key

## Expected Behavior
Window B should show "Task B prompt" (its own previous prompt).

## Actual Behavior
Window B may show "Task A prompt" (the prompt from Window A) or some other prompt from a different window/session.

## Possible Cause (Brief Analysis)

The issue appears to be related to how session identifiers are managed during session restoration:

1. Each window's prompt history is stored using a scoped key that includes the workspace directory and session ID
2. When a window is reopened and the session is restored, the session ID may not be correctly re-associated
3. This can cause the window to read from or write to a shared/default storage location that other windows also access
4. As a result, the prompt history becomes cross-contaminated between what should be isolated sessions

The storage key generation logic (likely in `Persist.scoped()` or similar) may be producing identical keys for different sessions when the session ID is undefined or not properly restored.

## Environment

- MiMo Code version: 0.1.0
- Operating System: Windows 11
- Installation method: `npm install -g @mimo-ai/cli`

## Additional Context

- The issue does not occur when both windows remain open simultaneously
- The issue specifically manifests after window close + reopen + session restore
- The cross-contamination appears to be unidirectional or random (not consistently bidirectional)

nightt5879:
## Summary
- store prompt history entries with session/workspace scope metadata
- filter Up/Down prompt history by active session, falling back to workspace scope only when no session is active
- keep legacy unscoped entries out of restored sessions so old global history cannot contaminate a resumed session

Fixes #392

## Tests
- `npx -y bun@1.3.11 test test/cli/tui/prompt-history.test.ts test/cli/tui/thread.test.ts`
- `npx -y bun@1.3.11 typecheck`
- `npx -y bun@1.3.11 test test/cli/tui` *(known Windows temp cleanup failures in plugin runtime tests: EBUSY removing temp fixture directories; prompt-history tests passed in this run)*


qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/992

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1069

PR #408, opened 2026-06-12
fix: prevent lazy() from permanently poisoning on first-call failure

xtyooo:
## 概述

`packages/shared/src/util/lazy.ts` 中的 `lazy()` 辅助函数在执行工厂函数**之前**就设置了 `loaded = true`。如果工厂函数首次调用抛出异常，所有后续调用都会静默返回 `undefined` 而不会重试。该函数在代码库中有 28 个消费者（数据库初始化、shell、provider、服务器路由等），一个瞬时的启动错误会永久性地损坏 lazy 值。

## 根因

```typescript
// 修复前（有 bug）
return (): T => {
  if (loaded) return value as T
  loaded = true    // 在 fn() 执行前就设置了
  value = fn()
  return value as T
}
```

如果 `fn()` 抛出异常，`loaded` 已经是 `true`，而 `value` 仍然是 `undefined`。后续每次调用都返回 `undefined as T`。

## 修复方案

将 `loaded = true` 移到 `value = fn()` **成功执行之后**：

```typescript
// 修复后
return (): T => {
  if (loaded) return value as T
  value = fn()      // 先执行
  loaded = true      // 成功后才标记为已加载
  return value as T
}
```

## 影响范围

- **涉及文件**: `packages/shared/src/util/lazy.ts`（28 个消费者）
- **风险**: 低 — 仅重新排列了两条赋值语句；正常路径无行为变化
- **是否破坏性**: 否

## 测试建议

- 验证首次调用工厂函数抛出异常后，可以成功重试
- 验证首次调用成功后仍然会缓存值（无回归）
- 运行: `cd packages/shared && bun test`


fengjikui:
I verified this against current `origin/main`.

On current main, the shared `lazy()` does get poisoned after a first-call throw. Minimal probe from `packages/shared`:

```bash
bun -e 'import { lazy } from "./src/util/lazy"; let calls = 0; const value = lazy(() => { calls++; if (calls === 1) throw new Error("transient"); return "ok" }); try { value() } catch {}; console.log(`second=${String(value())} calls=${calls}`)'
```

Current main output:

```text
second=undefined calls=1
```

After applying this PR diff, the same probe retries and succeeds:

```text
second=ok calls=2
```

I also ran the package checks after applying the PR:

```bash
cd packages/shared
bun test --timeout 30000
bun typecheck
git diff --check
```

All passed for me with Bun 1.3.14. The only thing I would still add is a focused regression test for this first-throw-then-retry case, since `packages/opencode` already has lazy tests but `packages/shared` does not currently cover this failure mode.

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/993

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1068

PR #429, opened 2026-06-12
Fix numpad input in prompt composer

### LINKED ISSUE #422: Numpad Keys Not Working in Prompt Input Field
## Description

The numeric keypad (numpad) keys do not work when typing in the prompt input field in MiMo Code. Pressing any key on the numpad (digits 0-9, decimal point, or arithmetic operators like +, -, *, /) produces no input in the text field.

## Expected Behavior

Numpad keys should function identically to the main keyboard number keys, inserting the corresponding digits and symbols into the prompt input field.

## Actual Behavior

No characters are inserted when pressing numpad keys. The input field remains unchanged.

## Reference: Other Applications Work Correctly

This issue appears to be specific to MiMo Code's input handling. The numpad works correctly in:

- **OpenCode** (the predecessor/fork base of MiMo Code) - numpad inputs work normally
- **VS Code** terminal and editor
- **Windows Notepad**, **Word**, and other native applications
- **Web browsers** (Chrome, Edge) in text input fields
- **Claude Code** and other CLI-based coding assistants

This suggests the issue lies in how MiMo Code's input field captures or processes keyboard events, specifically for numpad key codes.

## Steps to Reproduce

1. Open MiMo Code (Web UI or Desktop app)
2. Click on the prompt input field at the bottom
3. Ensure NumLock is enabled (numpad should be in number input mode)
4. Press any digit key on the numpad (0-9)
5. **Observe**: No digit appears in the input field
6. Try the main keyboard number row (above QWERTY)
7. **Observe**: These work correctly and digits appear

## Environment

- MiMo Code version: 0.1.0
- Operating System: Windows 11
- Installation method: npm install -g @mimo-ai/cli
- Hardware: Standard keyboard with numeric keypad
- NumLock state: ON (tried both ON and OFF)

## Possible Cause (Brief Analysis)

The issue likely stems from how the prompt input field handles keyboard events:

1. **Key Code Mapping**: MiMo Code may be using `event.key` or `event.code` for keyboard input handling. Numpad keys produce different values:
   - Main keyboard "1": `event.key = "1"`, `event.code = "Digit1"`
   - Numpad "1": `event.key = "1"`, `event.code = "Numpad1"`
   
   If the code checks for `event.code.startsWith("Digit")`, it would miss numpad inputs.

2. **ContentEditable Handling**: The prompt input appears to use a `contentEditable` div rather than a native `<input>` element. `contentEditable` elements have inconsistent behavior with numpad input across browsers and Electron-based apps, especially when combined with custom keyboard event interception.

3. **IME Composition Handling**: The code includes checks for `event.isComposing` and `event.keyCode === 229` (IME composing state). Numpad inputs during or after IME composition may be incorrectly filtered out.

4. **PreventDefault Interference**: The input field has extensive `handleKeyDown` event handling with many `event.preventDefault()` calls for special keys (Arrow keys, Enter, Escape, etc.). It's possible numpad key events are being inadvertently caught in this net, especially if the numpad Enter key is treated differently than the main Enter key.

The fact that OpenCode (the upstream/base project) does not have this issue suggests this is a regression or a specific change in MiMo Code's input handling implementation.

nightt5879:
## Summary

- Add a numpad key helper that normalizes printable `Numpad*` keydown events, including named and unidentified WebView values.
- Route printable numpad keys through the prompt input's existing `addPart` and `handleInput` path so digits/operators are inserted in the contenteditable composer.
- Cover digits, arithmetic operators, fallback key values, shortcuts, and NumLock navigation keys with unit tests.

Fixes #422.

## Validation

- `npx --yes bun@1.3.11 test ./src/components/prompt-input/numpad.test.ts` from `packages/app`
- `git diff --check`
- `bun typecheck` from `packages/app` was attempted and starts, but fails on the existing Windows checkout/symlink issue in `src/custom-elements.d.ts` line 1 (`../../ui/src/custom-elements.d.ts`).


fengjikui:
I applied this PR on current `origin/main` and verified the focused app-side checks locally from `packages/app`:

```bash
bun test ./src/components/prompt-input/numpad.test.ts --timeout 30000
bun typecheck
git diff --check
```

All passed for me with Bun 1.3.14. I also checked the insertion path: the new handler routes printable numpad keys through the existing `addPart({ type: "text" })` flow, which inserts at the current selection/range and then calls `handleInput()`, so it matches the prompt composer’s normal text-update path.

Scope note: this looks like good coverage for the Web/App prompt issue in #422. I would not treat it as covering #506 yet, since #506 reports Windows PowerShell / TUI behavior where numpad keys become arrow keys, which is likely a different input stack.

nightt5879:
Thanks for the detailed check! @fengjikui

Agreed — this PR only covers the Web/App prompt composer path for #422. I’m leaning toward keeping this as a focused fix and handling #506 as a separate follow-up PR, but happy to adjust if maintainers prefer expanding the scope here.

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/995

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1066

PR #431, opened 2026-06-12
fix: use startsWith for xiaomi providerID check in websearch tool

### LINKED ISSUE #430: fix: websearch tool does not route to xiaomi backend for xiaomi-token-plan-cn provider
## Bug Description

The websearch tool in `packages/opencode/src/tool/websearch/index.ts:66` checks `model?.providerID === "xiaomi"` to route to the xiaomi web search backend. However, the xiaomi provider can also be configured with providerID `xiaomi-token-plan-cn`, which fails this strict equality check and falls back to Exa.

## Steps to Reproduce

1. Use opencode with a xiaomi provider configured as `xiaomi-token-plan-cn`
2. Trigger a web search
3. The search routes to Exa instead of xiaomi's built-in `web_search` tool

## Expected Behavior

All xiaomi-prefixed providers should route to the xiaomi web search backend.

## Proposed Fix

Change the condition from strict equality to `startsWith`:

```ts
// Before
model?.providerID === "xiaomi"

// After
model?.providerID.startsWith("xiaomi")
```

ponponon:
## Problem

The websearch tool uses strict equality (`=== "xiaomi"`) to route to xiaomi's built-in web search backend. But xiaomi providers can also be configured with providerID `xiaomi-token-plan-cn`, which fails this check and incorrectly falls back to Exa.

## Fix

Changed `model?.providerID === "xiaomi"` to `model?.providerID.startsWith("xiaomi")` to match all xiaomi-prefixed providers.

Fixes #430


xiaokhkh (commented):
Finding: this fixes the routing check inside `WebSearchTool`, but the tool still appears to be gated out earlier for the same prefixed provider IDs.

`ToolRegistry.tools` only exposes `websearch` when `input.providerID === "xiaomi"` (or opencode / EXA flag). For `xiaomi-token-plan-cn`, the tool can still be filtered out before `WebSearchTool.execute` reaches the new `startsWith("xiaomi")` branch, so #430 may remain unfixed for the reported provider ID.

I would update the registry gate alongside this change, e.g. use the same xiaomi-prefix predicate there, and add a focused test that `websearch` is available for `xiaomi-token-plan-cn`.

fengjikui:
I applied this PR on current `origin/main` and added a temporary focused registry regression in a local worktree to isolate the provider gate from agent permissions:

```ts
for (const providerID of ["xiaomi", "xiaomi-token-plan-cn"]) {
  const tools = yield* registry.tools({
    providerID: ProviderID.make(providerID),
    modelID: ModelID.make("test-model"),
    agent: { permission: Permission.fromConfig({ "*": "allow" }) } as any,
  })
  console.log(providerID, tools.map((tool) => tool.id).filter((id) => id.includes("search") || id.includes("fetch")))
  expect(tools.map((tool) => tool.id)).toContain("websearch")
}
```

Result after applying this PR:

```text
xiaomi [ "webfetch", "websearch" ]
xiaomi-token-plan-cn [ "webfetch" ]
```

So the execution branch in `WebSearchTool` is fixed, but #430 still remains reachable because `ToolRegistry.tools()` filters out `websearch` before the tool can execute for `xiaomi-token-plan-cn`. The existing registry test suite still passes; this needs an additional registry-level regression plus the same xiaomi-prefix predicate at the gate in `packages/opencode/src/tool/registry.ts`.

ponponon:
Good catch! Added the same `startsWith` fix to `ToolRegistry.tools()` in `packages/opencode/src/tool/registry.ts:315`.

Both locations now use `input.providerID.startsWith("xiaomi")` to correctly gate the websearch tool for all xiaomi-prefixed providers.

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/996

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1065

PR #274, opened 2026-06-11
fix: startup crash in git repos (EEXIST) and unbounded log file growth

whitelonng:
## Summary

Fixes #243, fixes #209, fixes #219

Two reliability fixes for issues that make MiMo Code unusable for affected users:

### 1. Startup crash in git repositories (#243)

`setupProjectIdEnvironment` in `packages/opencode/src/project/project.ts` is a best-effort bookkeeping step, but it runs inside `Effect.promise()`, so **any** filesystem error there becomes a fatal startup crash (`UnknownError: EEXIST: file already exists, mkdir '...\.git\info'`). The reporter hit this on Windows, where `mkdir` with `recursive: true` can still throw EEXIST/EPERM on network/synced drives, and the same crash occurs whenever `.git/info` is not writable as a directory.

Fix: each filesystem step in this best-effort function now tolerates failure instead of killing startup.

### 2. Log directory grows without bound (#209, #219)

`Log.init` keeps the 10 newest log files but puts **no size limit on the current file** — one user reported ~200GB of logs in 8 hours, with the machine lagging while they were written. Whatever is doing the heavy logging, the logger should never be able to fill the disk.

Fix: rotate the active log file once it reaches 20MB (counter-suffixed filenames keep same-second rotations unique) and prune rotated files with the existing `cleanup()`, bounding the log directory to roughly `10 × 20MB`.

## Test plan

- [x] New test: `fromDirectory` no longer crashes when `.git/info` exists as a file (fails on `main`, passes with the fix)
- [x] New test: writing past the 20MB cap rotates to a new log file and the old file stays bounded
- [x] `bun test test/project/project.test.ts test/util/log.test.ts` — 30/30 pass
- [x] `bun typecheck` from `packages/opencode` — clean

PR #444, opened 2026-06-12
fix: catch EEXIST when .git/info already exists during project setup

### LINKED ISSUE #405: 在包含git仓库的目录运行 mimo 命令时遇到 EEXIST 报错
### Description

现象：在包含 Git 仓库的目录下运行 mimo 时会直接崩溃，报错内容如下：PS D:\onedrive\notebook> mimo
{
  "name": "UnknownError",
  "data": {
    "message": "Error: EEXIST: file already exists, mkdir 'D:\\onedrive\\notebook\\.git\\info'\n    at async KN (B:/~BUN/root/chunk-3k7cmhct.js:16:9936)"
  }
}
原因排查： mimo 在执行时试图创建 .git\info 目录，但是由于当前处于 Git 仓库中，该目录（通常包含 exclude 文件）已经存在。代码中在调用 mkdir 时似乎没有判断目录是否存在，也没有添加递归创建参数（例如 Node/Bun 中的 { recursive: true }），导致抛出 EEXIST 异常并中断运行。
建议修复： 在创建目录的代码处，加上对已存在目录的兼容处理（如捕获 EEXIST 错误或使用 { recursive: true }），避免在标准 Git 仓库中运行时崩溃。

### Plugins

没有使用任何插件

### OpenCode version

npm下载的最新版0.1.0

### Steps to reproduce

找一个已经建有git的odrive目录，在其目录下打开mimo code即可

Image

Image

### Operating System

win11

### Terminal

windows terminal使用powershell7

Code-suphub:
## Problem

On Windows, running `mimo` in a directory containing a git repository causes an immediate crash with:

```
Error: EEXIST: file already exists, mkdir 'D:\onedrive\notebook\.git\info'
    at async KN (B:/~BUN/root/chunk-3k7cmhct.js:16:9936)
```

Closes #405

## Root Cause

`setupProjectIdEnvironment` in `packages/opencode/src/project/project.ts` calls `nodeFs.mkdir()` with `{ recursive: true }` to ensure `.git/info` exists. However, Bun's Node.js compatibility layer on Windows can throw `EEXIST` even with `recursive: true` when the target directory already exists.

The error stack trace (`B:/~BUN/root/chunk-...`) confirms this is a Bun runtime compatibility issue.

## Fix

Wrap the `mkdir` call with a `.catch` that specifically ignores `EEXIST` errors while re-throwing anything else. This matches the defensive pattern already used elsewhere in the codebase (e.g., `flock.ts`).

## Changes

| File | Change |
|------|--------|
| `packages/opencode/src/project/project.ts` | Add EEXIST-safe catch on mkdir call |
| `packages/opencode/test/project/project.test.ts` | Add regression test for pre-existing .git/info |

## Testing

- All 28 project tests pass (including new regression test)
- New test `.git/info already exists (#405)` pre-creates the directory and verifies no crash

```
bun test test/project/project.test.ts
# 28 pass, 0 fail
```


fengjikui:
I applied this PR on current `origin/main` and the local focused checks pass from `packages/opencode`:

```bash
bun test test/project/project.test.ts -t "should not crash when .git/info already exists" --timeout 30000
bun typecheck
git diff --check
```

One caveat: on macOS/Linux this regression does not actually exercise the new `EEXIST` catch, because `fs.mkdir(path, { recursive: true })` succeeds when `.git/info` is already a directory. So the test verifies the normal pre-existing-directory path and the exclude update, but not the Windows/Bun failure mode reported in #405.

There is also duplicate history around this bug: #405 was called out as a duplicate of #243, with older PRs #260 and #274 already open. If this PR is the preferred path, I think it would be stronger with either a test that simulates `mkdir` throwing `EEXIST`, or a note that the regression is platform-limited and the local test only covers the non-crashing happy path.

PR #445, opened 2026-06-12
fix(tui): context meter shows summarize call usage right after compaction

### LINKED ISSUE #435: compact之后上下文反倒变多了
### Description

大约在40%上下文的时候执行了一次compact，执行完后变成97%了

### Plugins

No plugin

### OpenCode version

0.1.0

### Steps to reproduce

_No response_

Image

### Operating System

Ubuntu 24.04 LTS

### Terminal

VS Code 1.124.0-insider

fancyboi999:
## Description

Fixes #435

Right after a compaction finishes, the context meter jumps to near-full (the screenshot in #435 shows 971.0K / 97% after compacting at ~40%).

Cause: the meter reads the latest assistant message's full token usage (`input + output + reasoning + cache`). Right after compaction that message is the summary itself, and its input/cache tokens describe the *discarded* pre-compact history the summarize call consumed — not the new context. The backend already special-cases this (`processor.ts` skips the overflow check for summary messages, `prompt.ts` skips the memory-flush nudge), but the display sites don't.

## Changes

- New `contextTokens()` helper: for a `summary === true` message only `tokens.output` counts — the summary text is what seeds the new context window
- Applied in the TUI status bar, subagent footer, and sidebar context widget
- Same fix in the app's `session-context-metrics.ts`

## How I verified

- `bun test test/cli/tui/context-tokens.test.ts` (packages/opencode): 2 pass
- `bun test src/components/session/session-context-metrics.test.ts` (packages/app): 5 pass, incl. the new summary case
- `bun typecheck` clean in both packages
- Full `test/cli/tui/` suite: 98 pass + 1 pre-existing fail in `thread.test.ts` (unhandled error about `workflow/builtin/deep-research.js`; reproduces identically on an unmodified checkout)

Not reproduced on a live 1M-token session (needs ~400K of real history), but the math matches the report: the meter value right after compact equals the summarize call's own usage.

Note: upstream opencode's `prompt/index.tsx` has the same unguarded calculation, so this may be worth upstreaming.


qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/997

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1064

PR #468, opened 2026-06-12
fix: evict completed WorkflowRuntime entries from runs Map

### LINKED ISSUE #467: fix: WorkflowRuntime `runs` Map retains completed entries indefinitely (no eviction)
## Summary

`WorkflowRuntime` uses an in-memory `runs: Map<string, RunEntry>` to track workflow execution state. Entries are added on every `launch()` call but **never removed** — neither on completion, failure, nor cancellation. Each `RunEntry` holds a `Deferred`, `Fiber` reference, three `Set`s (`childActorIDs`, `worktrees`, `childRunIDs`), and other bookkeeping fields.

While individual entry size is modest (~few KB), the map grows monotonically for the lifetime of the process. For long-running instances that execute many workflows (CI/CD agents, session servers), this is a slow but unbounded leak.

## Code Location

`packages/opencode/src/workflow/runtime.ts`

```typescript
// L195 — declaration
const runs = new Map<string, RunEntry>()

// L399 — write (every launch)
runs.set(runID, entry)

// L1139 — resume also sets (overwrites stale terminal entry)
// comment: "runs.set overwrites the stale terminal entry"

// ❌ No runs.delete() / runs.clear() / eviction anywhere in the file
```

**`RunEntry` structure (L50-72):**
```typescript
interface RunEntry {
  runID: string
  sessionID: SessionID
  status: RunStatus                    // "running" | "completed" | "failed" | "cancelled"
  deferred: Deferred.Deferred<RunOutcome>
  fiber: Fiber.Fiber<void> | undefined
  childActorIDs: Set<string>
  worktrees: Set<string>
  childRunIDs: Set<string>
  name: string
  running: number
  succeeded: number
  failed: number
  agentCount: number
  capWarned: boolean
  warnedModelRefs: Set<string>
  currentPhase: string | undefined
}
```

## Steps to Reproduce

1. Write a test that starts multiple sequential workflow runs via `WorkflowRuntime.start()` + `WorkflowRuntime.wait()`
2. After all runs complete, observe that the internal `runs` Map retains every entry

**Reproduction test:**
```typescript
const runtime = yield* WorkflowRuntime.Service
const session = yield* Session.Service
const parent = yield* session.create({ title: "leak-test", permission: [...] })

for (let i = 0; i < 20; i++) {
  yield* llm.text("done")
  const script = `export const meta = { name: "run-${i}", description: "d" }`
  const { runID } = yield* runtime.start({ script, sessionID: parent.id, parentActorID: "main", model: ref })
  yield* runtime.wait({ runID })
}
// runs.size === 20 — all entries retained after completion
```

**Run with:**
```bash
cd packages/opencode && bun test test/workflow/runtime.test.ts
```

## Observed Behavior

```
Runs executed: 20
Heap delta:   +43.63 MB
RSS delta:    +53.95 MB

Runs | Heap (MB) | RSS (MB)
-----|-----------|---------
  10 |    115.74 | 333.72
  20 |    116.15 | 336.23
  30 |    116.49 | 338.64
```

Per-entry overhead is modest (~few KB), so this is not an acute OOM risk. However:
- The map grows monotonically with no upper bound
- Terminal entries (`completed`/`failed`/`cancelled`) are never reclaimed
- `resume()` further compounds the issue by overwriting stale entries without cleaning up the old fiber reference

## Expected Behavior

Completed/failed/cancelled runs should be evicted from the `runs` Map after a reasonable grace period, or capped at a maximum count.

## Suggested Fix

```typescript
// Option A: Delayed eviction (grace period for status/wait queries)
// In the terminal path (after Deferred.succeed):
setTimeout(() => runs.delete(runID), 30_000)

// Option B: Cap at N most recent runs
if (runs.size > MAX_RETAINED_RUNS) {
  const oldest = runs.keys().next().value
  runs.delete(oldest)
}
```

**Caution:** `status()`, `wait()`, and `cancel()` all do `runs.get(runID)`. Immediate deletion would break in-flight queries. Either use a grace period (Option A) or only evict terminal entries that are no longer `running` (Option B).

## Impact

- **Severity:** Low-Medium (slow leak, not acute OOM)
- **Trigger:** Any long-running process that executes many workflows
- **User-facing:** Gradual memory increase over hours/days of continuous use
- **Fix complexity:** Low (one eviction guard)
- **Regression risk:** Low (ensure `status`/`wait`/`cancel` still resolve within the eviction window)

## Validation

```bash
cd packages/opencode && bun test test/workflow/runtime.test.ts
cd packages/opencode && bun typecheck
```

bcoderg-sys:
## Summary

The `runs` Map in `WorkflowRuntime` only ever added entries via `set()` but never removed them. Each completed/failed `RunEntry` (holding a `Deferred`, `Fiber` reference, and multiple `Set`s) was retained for the lifetime of the process.

This change adds a 30-second delayed eviction after a run reaches terminal state (completed or failed), giving `status`/`wait`/`cancel` callers a grace period to read the outcome before the entry is removed.

Closes #467

## Changes

- `packages/opencode/src/workflow/runtime.ts`: Add `setTimeout(() => runs.delete(runID), 30_000)` in both the success and failure terminal paths
- `packages/opencode/test/workflow/runs-leak.test.ts`: Reproduction test that runs 30 sequential workflows and reports memory growth

## Testing

```bash
# Run the leak reproduction test
cd packages/opencode && bun test test/workflow/runs-leak.test.ts

# Run existing workflow tests (1 flaky cancel test pre-existing)
cd packages/opencode && bun test test/workflow/runtime.test.ts

# Typecheck
bun turbo typecheck
```

All existing tests pass. The one timeout failure (`cancel during an in-flight fan-out reclaims every child`) is pre-existing on `main` and unrelated to this change.

## Impact

- **Severity:** Low-Medium (slow leak, not acute OOM)
- **Fix complexity:** Low (2 lines of eviction logic)
- **Regression risk:** Low (30s grace period preserves all status/wait/cancel semantics)


xiaokhkh (commented):
Finding: cancelled runs still leak because this only schedules eviction from the success/failure work-fiber paths.

The issue and PR summary both describe reclaiming terminal entries, including `cancelled`, but `cancelEntry` sets `entry.status = "cancelled"` and succeeds the deferred without deleting or scheduling deletion of that `RunEntry`. Since `cancel()` calls `cancelEntry(entry)` directly, user-cancelled workflows still remain in the `runs` Map for the lifetime of the process.

I would schedule the same delayed `runs.delete(entry.runID)` in `cancelEntry` as well, and add/extend a cancel-path test so the leak fix covers all terminal states.

bcoderg-sys:
@xiaokhkh Good catch! Added `setTimeout(() => runs.delete(entry.runID), 30_000)` to `cancelEntry` as well.

fengjikui:
I rechecked the current PR after the cancel-path follow-up because this touches workflow lifecycle semantics.

Applied the PR diff on top of current `origin/main`, then ran the targeted tests plus package typecheck:

```sh
gh pr diff 468 --repo XiaomiMiMo/MiMo-Code --patch > /tmp/mimo-pr468.patch
git apply /tmp/mimo-pr468.patch
cd packages/opencode
PATH="$HOME/.bun/bin:$PATH" bun test test/workflow/runtime.test.ts -t "cancelled run is evicted|resume re-launches|list returns persisted runs" --timeout 70000
cd ../..
PATH="$HOME/.bun/bin:$PATH" bun run --cwd packages/opencode typecheck
```

Result:

```text
3 pass, 0 fail
packages/opencode typecheck: pass
```

The targeted run covers the newly added cancel eviction case and two nearby persistence/resume invariants. That is useful because terminal entries are evicted from the in-memory `runs` Map after the grace period, but `resume()` still has to work via `WorkflowPersistence` when the live entry is gone. The existing resume test still passes with this patch applied.

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/999

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1062

PR #477, opened 2026-06-12
fix: allow models from opencode/opencode-go providers

Ymaomi:


xiaokhkh (commented):
Finding: this PR needs to update the existing provider test or it will fail CI.

`test/provider/provider.test.ts` still asserts that `MimoFreeAuthPlugin` disables `opencode` and `opencode-go`, but this patch removes the code that pushes those providers into `disabled_providers`. Running the focused test on this branch fails with `Expected: false, Received: true` at the `opencodeProviderPresent(providers)` assertion.

Command run from `packages/opencode`:

```bash
bun test test/provider/provider.test.ts -t "opencode and opencode-go providers are disabled by MimoFreeAuthPlugin" --timeout 30000
```

If unblocking these providers is the intended behavior, the test should be updated to assert the new provider set and the PR description should call out the changed boundary between the free `mimo` provider and user-configured paid `opencode` / `opencode-go` providers.

Ymaomi:
@xiaokhkh Thanks for the review catching the test issue. Here's what changed:

- Test renamed to `"MimoFreeAuthPlugin hides opencode-zen provider by default"`. The `apiKey` was removed from the test config — providing a key now correctly activates opencode instead of being silently overridden.

- `disabled_providers` approach was wrong: the plugin's `config` hook runs before auth is loaded, so it could never see credentials from `/connect`. The hiding logic moved into the opencode custom loader in `provider.ts`, which checks `dep.auth()` alongside env and config — all three credential sources are recognized.

- The boundary: by default, only `mimo` free is visible and opencode-zen free models are hidden. `opencode-zen` and `opencode-go` appear when the user provides `OPENCODE_API_KEY`, sets `apiKey` in config, or authenticates via `/connect`.

屏幕截图 2026-06-14 111959

屏幕截图 2026-06-14 113426

屏幕截图 2026-06-14 113458

屏幕截图 2026-06-14 113555

屏幕截图 2026-06-14 113610

fengjikui:
I applied the latest diff on current `origin/main` and verified the focused provider behavior locally.

Checks that passed:

```bash
cd packages/opencode
bun test test/provider/provider.test.ts -t "MimoFreeAuthPlugin hides opencode-zen provider by default" --timeout 30000
git diff --check
```

I also probed the provider visibility boundary after applying the PR:

```text
empty config:        opencode=absent; opencode-go=absent; mimo=1
opencode apiKey:    opencode=45 models; opencode-go=absent; mimo=1
opencode-go apiKey: opencode=absent; opencode-go=13 models; mimo=1
both apiKey:        opencode=45 models; opencode-go=13 models; mimo=1
```

That matches the intended boundary from your follow-up comment: MiMo free stays visible by default, OpenCode providers are hidden without credentials, and user-configured `opencode` / `opencode-go` credentials make those providers available. I did not complete a full package typecheck in this local pass, so this is focused verification rather than full CI coverage.

Ymaomi:
**Upstream changes**

The upstream repo released a major update that restructured the plugin system:
- `MimoFreeAuthPlugin` (free JWT auth) was replaced by `MimoAuthPlugin` (OAuth login)
- Provider renamed from `mimo` to `xiaomi`
- The `mimo-free.ts` file was deleted and replaced by `mimo.ts`

This meant our previous approach (hiding opencode via `disabled_providers` in the plugin hook) had to be reworked against the new codebase.

**What we changed**

1. Removed the `disabled_providers` blocking logic from `mimo.ts` — this was the root cause preventing opencode from appearing even when credentials exist.

2. Moved the hiding logic into the opencode custom loader in `provider.ts`, which checks `dep.auth()` alongside env and config. The loader can now see credentials from `/connect`, so opencode appears correctly when the user authenticates.

3. Updated the test to match the new plugin name and removed stale assertions (the old `mimo-auto` model no longer exists in the new version).

**Regarding mimo free provider**

The new `MimoAuthPlugin` uses OAuth login and doesn't inject a free model provider by default. The free tier behavior is not currently replicated in the new version. This is an upstream decision — we just ensure opencode works correctly when the user adds credentials.

**Commit history**

The history was rewritten because I accidentally committed a file with private data during the merge. I reset the branch to keep only the essential changes clean.

## Xiaomi(mimo)

屏幕截图 2026-06-16 024702

## Opencode-go
屏幕截图 2026-06-16 024421

## Opencode-zen(Too many models, showing free ones only)
屏幕截图 2026-06-16 024436

PR #505, opened 2026-06-13
fix: restrict bash in plan mode

### LINKED ISSUE #475: BUG Plan mode allows filesystem modifications via Bash tool
### Description

When plan mode is active, the agent can execute filesystem-modifying commands through the bash tool (e.g., mkdir, touch, file creation). According to the documentation, plan mode should be a "Read-only analysis mode for code exploration and solution design", but this restriction is not enforced on bash commands.
Steps to Reproduce
1. Enter plan mode (Tab → plan)
2. Ask the agent to create a folder: mkdir test
3. The folder is created successfully despite plan mode being active
Expected Behavior
Plan mode should block all filesystem-modifying operations, including:
- bash commands: mkdir, touch, cp, mv, rm, redirects (>, >>)
- edit / write tools (already blocked)
Actual Behavior
Only edit and write tools are restricted in plan mode. The bash tool can execute any command, including destructive ones like mkdir, rm -rf, etc.
Suggested Fix
Expand plan mode restrictions to include filesystem-modifying bash commands. Options:
1. Deny all bash commands in plan mode (simplest, but may be too restrictive)
2. Add a denylist for destructive bash commands (mkdir, touch, rm, cp, mv, redirects, npm install, etc.)
3. Add a readonly flag to bash tool that blocks write operations
Environment
- OS: Windows 11
- Shell: PowerShell
- MiMo Code version: 0.1.0
- Model: mimo-auto
Related
- README states: "plan — Read-only analysis mode for code exploration and solution design"

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

Windows 11

### Terminal

PowerShell

cnYui:
### Issue for this PR

Closes #475

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Plan mode is documented as read-only, but the built-in plan agent did not set bash permissions, so bash inherited the default allow behavior. That left commands like `mkdir`, `touch`, and shell redirects able to modify the workspace.

This PR makes plan-mode bash default to deny and explicitly allows common read-only exploration commands: `pwd`, `ls`, `cat`, `grep`, `rg`, `find`, `git status`, `git diff`, `git log`, `git show`, `Get-ChildItem`, and `Get-Content`. User-defined permission overrides still flow through the existing config merge path.

Passed locally:

- `bun test test/agent/agent.test.ts -t "plan agent" --timeout 30000`
- `bun test test/tool/bash.test.ts -t "matches redirects in permission pattern" --timeout 30000`
- `bun test test/tool/bash.test.ts -t "always pattern has space before wildcard" --timeout 30000`
- `git diff --check`

Local limitation:

- `bun typecheck` currently fails before reaching this change with `src/storage/db.node.ts(1,30): error TS2307: Cannot find module 'node:sqlite' or its corresponding type declarations.`

### Screenshots / recordings

No UI change.

### Checklist


xiaokhkh (commented):
Finding: this still allows write-capable bash commands in plan mode.

The new plan rules allow broad patterns like `cat *`, `grep *`, `find *`, and `git diff *` after the `*` deny in `packages/opencode/src/agent/agent.ts`. Permission matching is done against the full bash command string, so these evaluate as `allow`:

- `cat package.json > out.txt`
- `grep foo file > out.txt`
- `find . -delete`
- `git diff --output=out.patch`

That keeps the core failure mode from #475 reachable through commands that look read-only at the executable level but can still mutate the workspace through redirects or flags. I would add negative tests for redirects/destructive flags and either deny those more-specific patterns or narrow the allowlist so plan mode only permits truly read-only forms.

cnYui:
Fixed in b346994.

What changed:
- added negative coverage for plan-mode bash redirects, `find -delete`, `find -exec`, and `git diff --output`
- added more-specific deny rules after the read-only allowlist so those write-capable forms no longer resolve to `allow`

Verified:
- `bun test test/agent/agent.test.ts -t "plan agent allows only read-only bash commands by default" --timeout 30000`
- `git diff --check`

Also checked the wider local signals: `bun test test/agent/agent.test.ts --timeout 30000` still has the existing unrelated `general agent denies todo tools` failure, and `bun run typecheck` still stops at `src/storage/db.node.ts` because this local TypeScript setup cannot resolve `node:sqlite`.

fengjikui:
I re-checked the current PR rules against the existing `Permission.evaluate` / `Wildcard.match` behavior. The latest update does fix the redirect examples from the review, but I think a few write-capable forms still resolve to `allow` because `find *` / `git diff *` remain broad allow rules:

```text
find . -execdir rm {} \;  => allow via find *
find . -delete -print      => allow via find *
git diff --output out.patch => allow via git diff *
```

I also verified in a temp repo that `git diff --output out.patch` does create a file.

So before this closes #475, it may need a few more negative tests and deny patterns, for example `find * -execdir *`, `find * -delete *`, and `git diff --output *` in addition to the existing `--output=*` forms. The larger question is whether string-pattern allowlisting can stay robust enough for plan mode, but these three cases are small enough to cover in this PR.

cnYui:
Fixed in ff901bc.

What changed:
- added negative coverage for `find . -execdir rm {} \;`, `find . -delete -print`, and `git diff --output out.patch`
- added deny rules for `find * -execdir *`, `find * -delete *`, `git diff --output *`, and `git diff * --output *`

Verified:
- `bun test test/agent/agent.test.ts -t "plan agent allows only read-only bash commands by default" --timeout 30000`
- `git diff --check`

Wider local checks are unchanged: `bun test test/agent/agent.test.ts --timeout 30000` still only fails on the existing unrelated `general agent denies todo tools` assertion, and `bun run typecheck` still stops on the existing local `node:sqlite` type resolution issue.

fengjikui:
Thanks for the quick follow-up. I applied the latest PR diff on current `origin/main`; the updated focused test passes locally:

```bash
bun test test/agent/agent.test.ts -t "plan agent allows only read-only bash commands by default" --timeout 30000
```

I found one more class that still resolves to `allow`: pipelines from an allowed read command into a write/destructive command. With the current rules, these all match the broad read allow rules (`cat *`, `rg *`, `git diff *`, `find *`, `Get-Content *`):

```text
cat package.json | tee out.txt              => allow
rg needle src | tee out.txt                 => allow
git diff | tee out.patch                    => allow
find . -print0 | xargs rm -f                => allow
Get-Content package.json | Set-Content out.txt => allow
```

The first, second, third, and fifth create/overwrite files; the `find | xargs rm` form can delete files. Since #475 is specifically about plan mode being read-only, I think this PR still needs negative coverage for pipe forms before closing the issue. A conservative fix would be to deny pipeline operators in plan-mode bash (`*|*` / `* | *`) unless there is a parsed-command allowlist for the command on the right-hand side.

cnYui:
Fixed in 115865a.

What changed:
- added negative coverage for the pipeline cases listed here: `cat | tee`, `rg | tee`, `git diff | tee`, `find | xargs rm`, and `Get-Content | Set-Content`
- added a conservative plan-mode bash deny rule for `*|*`, so pipelines from read-allowed commands no longer resolve through the broad read allowlist

Verified:
- `bun test test/agent/agent.test.ts -t "plan agent allows only read-only bash commands by default" --timeout 30000`
- `git diff --check`

Wider local checks are unchanged: `bun test test/agent/agent.test.ts --timeout 30000` still only fails on the existing unrelated `general agent denies todo tools` assertion, and `bun run typecheck` still stops on the existing local `node:sqlite` type resolution issue.

fengjikui:
I re-applied the latest PR diff (`115865a`) on current `origin/main` and confirmed the updated focused test still passes locally:

```bash
bun test test/agent/agent.test.ts -t "plan agent allows only read-only bash commands by default" --timeout 30000
```

The new pipeline deny rule covers the cases from my previous comment now:

```text
cat package.json | tee out.txt                    => deny
rg needle src | tee out.txt                       => deny
git diff | tee out.patch                          => deny
find . -print0 | xargs rm -f                      => deny
Get-Content package.json | Set-Content out.txt    => deny
```

I found one remaining bypass class with shell control operators / command substitution. These still resolve to `allow` because the full command starts from a broad read-allowed pattern:

```text
cat package.json; touch out.txt                   => allow
ls && touch out.txt                               => allow
git status && rm -rf x                            => allow
Get-Content package.json; Set-Content out.txt     => allow
cat package.json `touch out.txt`                  => allow
cat package.json $(touch out.txt)                 => allow
```

So I think #475 still needs negative coverage for `;`, `&&`, backticks, and `$()` before this PR closes the read-only plan-mode gap. A conservative pattern-level fix could deny shell control/substitution tokens in plan-mode bash (`*;*`, `*&&*`, `*\`*`, `*$(*`) unless/until command parsing is introduced.

cnYui:
Fixed in f59c3c8.

What changed:
- added negative coverage for the control/substitution cases listed here: `cat; touch`, `ls && touch`, `git status && rm -rf`, `Get-Content; Set-Content`, backtick substitution, and `$()` substitution
- added conservative plan-mode bash deny rules for `*;*`, `*&&*`, ``*`*``, and `*$(*`

Verified:
- `bun test test/agent/agent.test.ts -t "plan agent allows only read-only bash commands by default" --timeout 30000`
- `git diff --check`

Wider local checks are unchanged: `bun test test/agent/agent.test.ts --timeout 30000` still only fails on the existing unrelated `general agent denies todo tools` assertion, and `bun run typecheck` still stops on the existing local `node:sqlite` type resolution issue.

fengjikui:
I re-applied the latest PR diff (`f59c3c8`) on current `origin/main`. The updated focused test passes locally:

```bash
bun test test/agent/agent.test.ts -t "plan agent allows only read-only bash commands by default" --timeout 30000
```

I also replayed the previous bypass matrix. The cases from my last comment now resolve to `deny`:

```text
cat package.json | tee out.txt                 => deny
rg needle src | tee out.txt                    => deny
git diff | tee out.patch                       => deny
find . -print0 | xargs rm -f                   => deny
Get-Content package.json | Set-Content out.txt => deny
cat package.json; touch out.txt                => deny
ls && touch out.txt                            => deny
git status && rm -rf x                         => deny
Get-Content package.json; Set-Content out.txt  => deny
cat package.json `touch out.txt`               => deny
cat package.json $(touch out.txt)              => deny
```

I found two adjacent shell separators that still resolve to `allow`:

```text
cat package.json\ntouch out.txt => allow
cat package.json & touch out.txt => allow
```

Both can still run a write command after a read-allowed command in plan mode. If this PR stays with pattern-level hardening, I think it should also add negative coverage for newline command separators and single `&` background separators, for example `*\n*` / `*&*` in addition to the existing `*&&*` rule.

cnYui:
Fixed in 5343042.

What changed:
- added negative coverage for newline command separators and single `&` background separators
- added conservative plan-mode bash deny rules for `*\n*` and `*&*`

Verified:
- `bun test test/agent/agent.test.ts -t "plan agent allows only read-only bash commands by default" --timeout 30000`
- `git diff --check`

Wider local checks are unchanged: `bun test test/agent/agent.test.ts --timeout 30000` still only fails on the existing unrelated `general agent denies todo tools` assertion, and `bun run typecheck` still stops on the existing local `node:sqlite` type resolution issue.

fengjikui:
I re-applied the latest PR diff (`5343042`) on current `origin/main` and the focused test still passes locally from `packages/opencode`:

```bash
bun test test/agent/agent.test.ts -t "plan agent allows only read-only bash commands by default" --timeout 30000
```

The newly added newline and single-`&` cases are covered now. I found one adjacent bash execution form that still resolves to `allow`: process substitution.

```text
cat <(touch out.txt)              => allow
cat package.json <(touch out.txt) => allow
cat < <(touch out.txt)            => allow
cat <<< $(touch out.txt)          => deny
cat package.json || touch out.txt => deny
```

I also verified in a temp shell that `cat <(touch out.txt)` creates `out.txt`, so this is still a write-capable command hidden under a read-allowed `cat *` pattern.

If this PR continues with conservative pattern-level hardening, I think it needs negative coverage for process substitution too, e.g. `*<(*`, before closing #475. Longer term, this class of bypasses is a good argument for parsing shell commands instead of growing a token denylist, but this specific case is small enough to cover here.

cnYui:
Fixed in 4264ce0.

What changed:
- added negative coverage for the process substitution cases listed here: `cat <(touch out.txt)`, `cat package.json <(touch out.txt)`, and `cat < <(touch out.txt)`
- added a conservative plan-mode bash deny rule for `*<(*`, so those forms no longer resolve through the broad `cat *` allow rule

Verified:
- `bun test test/agent/agent.test.ts -t "plan agent allows only read-only bash commands by default" --timeout 30000`
- `git diff --check`

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1001

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1060

Issue #561, opened 2026-06-14
(bug) Subagent delegation often fails because tool call format is incorrect (agent uses string instead of JSON object)


Zireael:
### Description

The actor tool delegation often fails because tool requires the operation parameter to be a JSON object, not a string.

### Plugins

N/A

### OpenCode version

mimo code 0.1.0

### Steps to reproduce

Get the agent to attempt delegate tasks

### Screenshot and/or share link

<img width="773" height="374" alt="Image" src="https://github.com/user-attachments/assets/c243305e-7ec6-4e96-a646-bdc63dab375a" />

### Operating System

windows 11

### Terminal

windows terminal

Issue #529, opened 2026-06-13
Can't add custom provider


aaronnewsome:
### Description

I enter

* Name
* Display Name
* Base URL

all of those work fine, but when it asks for the key, nothing works. My provider is local so I don't need a key. Blank key can't continue. 'NONE' string entered, can't continue. A real key entered and can't continue. 

Where is the models file so I can just add my own provider and model.

### Operating System

Ubuntu 24.04

### Terminal

LX Terminal

aaronnewsome:
Figured it out. The dialog doesn't allow blank key. Once you try to proceed with a blank key, you can't provide one at that point and you need to start over. Entering NONE on the first try does allow the dialog to proceed.

The UI is extra unhelpful by not explaining what's happening and instead just freezing and refusing to continue.

Issue #487, opened 2026-06-13
模型列表无法搜索


hub-rb:
### Description

/models 切换模型无法进行搜索

zhumengzhu:
Hi, I can reproduce this. Adding screenshots to help clarify:

**MiMo-Code** — no search/filter input in `/models`:

<img width="994" height="860" alt="MiMo-Code: no search input" src="https://github.com/user-attachments/assets/c50f16ae-ffab-4a36-9963-2603fec6075f" />

**OpenCode** — has a working search input for reference:

<img width="994" height="242" alt="OpenCode: search input working" src="https://github.com/user-attachments/assets/5e938415-00a3-4b4a-8d4e-72b27061e569" />

**Root cause**: `DialogModel` passes `skipFilter={true}` to `DialogSelect`, which hides the filter input entirely. The custom `fuzzysort` block in `options()` never gets a chance to run because there's no input field rendered for the user to type into.

PR **#649** takes the right approach — removing the custom fuzzysort and letting `DialogSelect` handle filtering with its built-in support. One minor thing to note: the custom block searched both `title` and `category` (provider name), while the built-in only searches `title`. So searching by provider name (e.g. "openai" to filter OpenAI models) would be lost. Might be worth considering whether `DialogSelect` should also search the description/category field, or keeping `category` in the search scope.

PR #570, opened 2026-06-14
fix: actor tool stringified ops, model search, custom provider UX, editor cursor bug

MrRealORG:
Second round of bug fixes and UX improvements for MiMo-Code.

## Bug Fixes

- **#561**: Actor tool now handles stringified operation param via Zod union transform
- **#487**: Model search more lenient (fuzzysort threshold -6000 to -10000) for CJK support
- **#529**: API key field clearly marked optional for local providers (EN + ZH i18n)
- Editor cursor position bug fix: early return when remaining=0 on pill/break
- CLI: Truncate long busy status messages to 60 chars


fengjikui:
For the #561 actor-tool part: the normalization direction makes sense, but this PR currently bundles several unrelated changes and I do not see a focused regression for stringified `operation`. Since #561 is a tool-call contract bug, a small test around `packages/opencode/src/tool/actor.ts` would make this much safer to review.

One implementation note: the repo style guide prefers avoiding `try/catch` where possible. If the string parse branch stays, it may be worth keeping that change isolated and tested so reviewers can judge the compatibility tradeoff separately from model search/status bar/provider UI changes.

Issue #565, opened 2026-06-14
[FEATURE]: i18n — hardcoded Chinese strings in login plugin, audit all messages for English equivalents


HaleTom:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

**Summary:** The MiMo auth plugin (`packages/opencode/src/plugin/mimo.ts`) contains hardcoded Chinese strings that bypass the existing i18n system. These should be routed through the TUI i18n framework so all 17 supported locales display localized text.

**Context:** The TUI already has a mature i18n system supporting 17 locales (en, zh, zht, ko, de, es, fr, da, ja, pl, ru, ar, no, br, th, bs, tr) via `packages/opencode/src/cli/cmd/tui/i18n/`. Corresponding i18n keys already exist for the login flow — they're just not wired into the plugin.

**Hardcoded Chinese found:**

1. **`packages/opencode/src/plugin/mimo.ts`** — login method label:
   ```ts
   label: "浏览器登录",  // "Browser Login"
   ```
   This renders as Chinese regardless of the user's locale setting.

2. **`packages/opencode/src/plugin/mimo.ts`** — authorization instructions:
   ```ts
   instructions: "在浏览器中完成授权，或粘贴 Code 完成登录。",
   // "Complete authorization in the browser, or paste the Code to finish login."
   ```

3. **`packages/opencode/test/plugin/mimo.test.ts`** — test asserts the hardcoded Chinese:
   ```ts
   expect(hooks.auth!.methods[0].label).toBe("浏览器登录")
   ```

**Existing i18n keys that should be used:**

The TUI locale files already define these keys in `zh.ts`:
- `tui.dialog.login.flow.title` → "MiMo 登录"
- `tui.dialog.login.flow.placeholder` → "粘贴 Code（或等待浏览器回调）"
- `tui.dialog.login.flow.waiting` → "等待浏览器授权中..."
- `tui.dialog.login.flow.manual_hint` → "浏览器未打开？手动访问："
- `tui.dialog.login.flow.busy` → "登录中..."

And the corresponding `en.ts` keys exist as well. The plugin just needs to use them instead of hardcoded strings.

**Request:**
1. Replace the hardcoded Chinese strings in `mimo.ts` with i18n-aware lookups (the keys already exist).
[...]

Issue #502, opened 2026-06-13
响应的完整内容会输出到状态栏中


sui-an:

sui-an:
<img width="773" height="804" alt="Image" src="https://github.com/user-attachments/assets/1846bd44-2263-4490-962a-b382a8bfe459" />

PR #571, opened 2026-06-14
fix: i18n login plugin, model search, status bar overflow, paste pollution

MrRealORG:
## Summary

Change `detached` default for child process spawning to `true` on all platforms.

---

### Change

- `packages/opencode/src/util/process.ts`: Default `detached` to `true` instead of `!win32`, preventing child processes from taking over the parent console on Windows.

### Note

This PR was originally described as fixing #565, #487, #502, #504 but the actual diff only touches process spawning. Those issues will be addressed in separate PRs.


fengjikui:
The PR body says this closes #565, #487, #502, and #504, but the current diff only changes `packages/opencode/src/util/process.ts` by adding `detached: process.platform !== "win32"`.

That does not appear to touch:

- #565 i18n login strings
- #487 model dialog search
- #502 busy/status bar truncation
- #504 paste text pollution

This may still be useful for a process/terminal issue, but I would avoid closing those four issues from this PR unless the diff is updated to include the described fixes and focused verification.

MrRealORG:
You're right — the PR body was overpromising. I've updated it to accurately reflect the actual diff (only the `detached` default change in `process.ts`). Removed the Closes links for #565, #487, #502, #504. Those will be addressed in separate focused PRs.

fengjikui:
Thanks, confirmed the PR body now matches the actual diff and no longer closes #565 / #487 / #502 / #504. That makes the process-spawn change much easier to review on its own.

Issue #579, opened 2026-06-14
mimo fired over 50 api request in a sec after i used copy paste,


ji-podhead:
### Description

<img width="919" height="737" alt="Image" src="https://github.com/user-attachments/assets/3c6c57d2-1378-4b1e-99eb-8cd28027167f" />

mimo fired over 50 api request in a sec after i used copy paste, but the content is different from my clipboard:

```
s:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbf
fb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:
1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 9613
11cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbf
fb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cb
ebbffb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d
.js:1postMessagei @ 961311cbebbffb6d.js:1O @ 961311cbebbffb6d.js:1postMessagei @ 961311cbebbffb
[...]

Issue #578, opened 2026-06-14
Agent loop stops after the first tool call with local OpenAI-compatible models (Gemma via oMLX) — provider returns valid follow-up tool_calls


vitaliytv:
### Description

When using a **local OpenAI-compatible provider** (oMLX serving `gemma-4-e4b-it-OptiQ-4bit`), the agent loop **terminates after the first tool call**. `mimo run` on a multi-step task (e.g. "fix all lint errors") executes exactly **one** tool call (running the linter), receives the result, then ends the turn emitting a literal `<eos>` token as text — with **zero edits performed**.

Crucially, the model/provider is **not** at fault: when the same endpoint is sent a clean multi-turn conversation by hand (assistant `tool_call` + `tool` result), it correctly returns a **follow-up `tool_calls`** (an `edit` call) with `finish_reason: "tool_calls"`. So the model is fully capable of continuing the tool loop — mimo appears to drop it.

This makes mimo effectively unusable for agentic coding with this (and likely other Gemma-family) local model. Reproduced with both the default `build` agent and a custom agent.

### Observed mimo output

```
> build · gemma-4-e4b-it-OptiQ-4bit

I will run `oxlint` to identify all lint errors and then iteratively edit the specified
files to resolve every reported error, finally running `oxlint` again for verification.
<eos>
$ oxlint
run/.../comment.js:31:26: error unicorn(no-await-expression-member): ...
run/.../comment.js:60:28: error unicorn(no-await-expression-member): ...
run/.../comment.js:15:11: error unicorn(no-useless-spread): ...
<eos>
```

→ turn ends here, **0 files edited**. Note the literal `<eos>` (Gemma's end-of-turn token) leaking into the assistant content.

### Evidence the model continues correctly on a clean conversation

Direct `curl` to the exact same provider endpoint, giving the model an assistant `tool_call` followed by a `tool` result, then asking it to continue:

```bash
curl -s http://127.0.0.1:8000/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" -d '{
  "model": "gemma-4-e4b-it-OptiQ-4bit",
  "messages": [
    {"role": "system", "content": "You are a coding agent. After a tool returns, CONTINUE by calling the next tool until done."},
    {"role": "user", "content": "Fix the lint error in foo.js by editing it."},
    {"role": "assistant", "content": "", "tool_calls": [
[...]

PR #582, opened 2026-06-14
fix: 5 bug fixes — context default, plan Q&A, IME contrast, paste flood, EOS tokens

### LINKED ISSUE #534: Local Models Do Not Have Context Limits Defined, Causing Incorrect Context Usage Display and Potential Overflow
### Description

# Local Models Do Not Have Context Limits Defined, Causing Incorrect Context Usage Display and Potential Overflow

## Description

When using local OpenAI-compatible providers (llama.cpp in my case), MiMoCode does not appear to automatically know the model's actual context window.

As a result, the context usage indicator can be significantly incorrect, and context management may not trigger appropriately before the backend limit is reached.

## Environment

* MiMoCode: current release
* Provider: OpenAI-compatible (llama.cpp server)
* Backend context size: 180224 tokens
* Model: Qwen3.6-35B-A3B-REAM-192-Q4_K_M

## Observed Behavior

After configuring a local model without a `limit` section:

```json
"Qwen3.6-35B-A3B-REAM-192-Q4_K_M": {
  "name": "Qwen3.6-35B-A3B-REAM-192-Q4_K_M"
}
```

MiMo displayed:

```text
Context
22026 tokens
2% used
```

However, the backend context window was only:

```text
180224 tokens
```

22k tokens is approximately 12% of 180k, not 2%.

Later in a long-running coding session, the backend failed with:

```text
request (181631 tokens) exceeds the available context size (180224 tokens)
```

The agent repeatedly retried the request and became stuck.

## Investigation

I manually added limits to the local model definition:

```json
"Qwen3.6-35B-A3B-REAM-192-Q4_K_M": {
  "name": "Qwen3.6-35B-A3B-REAM-192-Q4_K_M",
  "limit": {
    "context": 180224,
    "output": 32000
  }
}
```

After restarting MiMo, the UI changed to:

```text
Context
20958 tokens
12% used
```

which correctly matches:

```text
20958 / 180224 ≈ 11.6%
```

This suggests MiMo is using the configured `limit.context` value for context accounting.

## Expected Behavior

For local/OpenAI-compatible models:

* MiMo should either detect the model context window automatically, or
* provide a documented way to configure it.

Additionally, intelligent context management/checkpointing should use the actual backend context size rather than an assumed/default value.

## Questions

1. Is `limit.context` the recommended way to configure local model context windows?
2. Does intelligent context management use `limit.context` when deciding when to checkpoint/reconstruct context?
3. Is there a default context size used when a local model does not define a limit?
4. Could the missing context limit be responsible for context-overflow situations where the backend rejects requests before MiMo performs reconstruction/compaction?

## Reproduction

1. Configure a local OpenAI-compatible model without a `limit` section.
2. Start a long-running coding session.
3. Observe context percentage in the UI.
4. Continue until backend context is exhausted.
5. Compare behavior with and without:

```json
"limit": {
  "context": 180224,
  "output": 32000
}
```

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

MrRealORG:
Fixes #534 #531 #518
Related #579 #578

- Lower DEFAULT_CONTEXT_WINDOW from 1M to 128K (#534)
- Skip plan workflow for simple Q&A messages (#531)
- Add IME composition CSS for light mode (#518)
- Suppress onInput during paste to prevent cascade (Related #579 — needs TUI path verification)
- Strip EOS tokens from text-delta for local models (Related #578 — needs deeper fix)

Note: #579 and #578 are marked as Related per reviewer feedback. The current fixes may be partial — these will be addressed in focused follow-up PRs with proper verification.


fengjikui:
Thanks for trying to cover several fresh reports here. Two of the linked issues may need a bit more evidence before this PR closes them:

- #579 reports Ubuntu + VS Code terminal. This patch changes `packages/app/src/components/prompt-input.tsx`, which looks like the web/app prompt path, not the TUI prompt path (`packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx`). Unless the reporter was using the web UI, this probably does not cover the reported paste flood.
- #578 may not be fixed by stripping literal EOS text alone. I tested current `main` with a fake OpenAI-compatible SSE provider returning `content` containing `<eos>` plus a normal `tool_calls` finish, and MiMo continued to the next tool call correctly. The failing oMLX/Gemma path likely depends on the exact streamed chunk/tool_call delta or finish_reason shape.

Suggestion: avoid `Fixes #579 #578` for now, or add focused regressions that exercise the exact TUI paste path and the Gemma/oMLX stream shape. Otherwise GitHub may auto-close two issues that are still unverified.

MrRealORG:
welcome lets do it

fengjikui:
Happy to help. I would split this into two reviewable tracks so maintainers can merge the verified parts without accidentally closing the unverified reports.

Suggested immediate PR body change for this PR:

- keep `Fixes #534 #531 #518`
- change `#578 #579` to `Related #578 #579` until there are focused reproductions

Then handle the two harder bugs separately:

1. **#579 paste flood**
   - The report says Ubuntu + VS Code terminal, so the likely target is the TUI prompt path: `packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx`.
   - A focused test/repro should show whether paste alone can submit/send, or whether it only pollutes input text.
   - The web `packages/app/src/components/prompt-input.tsx` change may still be useful, but it should not close #579 unless the reporter confirms they used the web UI.

2. **#578 Gemma/oMLX loop stop**
   - I would not treat literal EOS text as the whole root cause yet. My fake OpenAI-compatible SSE server returned `<eos>` plus a normal `tool_calls` finish and current main continued the agent loop correctly.
   - The useful regression target is the exact oMLX/Gemma streaming shape: tool-call deltas, `finish_reason`, and the follow-up request body after the first tool result.
   - If you can capture or mock that stream shape, I can help review whether the fix belongs in provider chunk normalization, session processor classification, or history/message conversion.

That keeps this PR mergeable for the three clearer fixes while leaving #578/#579 open for evidence-based follow-ups.

MrRealORG:
Updated the PR body per your suggestion:

- `Fixes #534 #531 #518` kept as-is (verified changes)
- `#579` and `#578` changed to `Related` with a note that these are partial fixes needing follow-up PRs

You're right that #579 needs the TUI path (`packages/opencode/.../prompt/index.tsx`), not the web app path. I'll create a focused PR for that separately.

fengjikui:
Thanks, confirmed the PR body now keeps `Fixes #534 #531 #518` and marks #579 / #578 as `Related` with follow-up notes. That should prevent GitHub from auto-closing the two unverified reports while keeping this batch reviewable for the clearer fixes.

I will watch for the focused TUI paste follow-up for #579, and for #578 I still think the next useful artifact is a raw oMLX/Gemma stream shape or a mock that matches it.

PR #586, opened 2026-06-14
fix: screen/tmux keyboard fix, status bar overflow, docx file reading

### LINKED ISSUE #526: I use screen , then in screen when to open mimo, I cannot type !
### Description

linux screen下无法工作。

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

MrRealORG:
Fixes #526 #502 #513

- #526: Detect screen/tmux in isPlainTerminal() so Kitty keyboard protocol is disabled, fixing ! key and other input issues
- #502: Truncate busy status messages to 80 chars to prevent response text overflowing into status bar
- #513: Add .docx text extraction via system unzip + XML tag stripping so the read tool can extract text from Word documents


fengjikui:
Thanks for looking at these. I think the #526 part is directionally right, but this PR may be easier to review if it is split by issue. A few concrete notes:

- #526: the screen/tmux detection should probably have a focused regression test for `TERM=screen*`, `STY`, `TERM=tmux*`, `TMUX`, and the explicit `MIMOCODE_TUI_PLAIN=false` opt-out.
- #502: status truncation is also covered in #570, so one of these PRs will likely become duplicate.
- #513: the `.docx` path is higher-risk than the TUI fixes. `execSync(`unzip -p "${filepath}" ...`)` goes through a shell and can break or become unsafe for paths containing quotes. It also only handles `.docx`, while the issue text mentions PDF too. If this stays, it probably needs either `spawn/execFile`-style argument passing or a zip/XML parser approach plus a focused fixture test.

My suggestion: keep the terminal keyboard fix as a small #526-only PR first, with tests, and move `.docx` reading into a separate PR once the exact file support scope is clear.

PR #589, opened 2026-06-14
fix(windows): use detached process to prevent child from taking over console

### LINKED ISSUE #544: UI becomes blank after "Run in background" for pygame application, session remains alive
### Description

# Bug: UI becomes blank after "Run in background" for pygame application, session remains alive

## Environment

* MiMo Code: v0.1.x
* OS: Windows 11
* Local model provider: OpenAI-compatible local endpoint (LM Studio / llama.cpp compatible)
* Python: 3.13.11
* Pygame: 2.6.1

## Summary

While testing a pygame application, MiMo reported:

```
[Interactive] Run SuperSnake game in background
$ python -m snake_game.main
```

After this, the MiMo dashboard disappeared and the terminal was taken over.

When reopening MiMo using:

```powershell
mimo -c
```

a blank MiMo window appeared with no visible session content.

However, the MiMo process, Python process, and session memory artifacts all remained alive.

This suggests the session was not lost, but the UI may have entered an unrecoverable state.

## Expected Behavior

* MiMo launches the game in the background.
* MiMo dashboard remains usable.
* If the dashboard is interrupted, reopening the session should restore the UI.

## Actual Behavior

* Dashboard disappears.
* Terminal shows:

```
pygame 2.6.1 (SDL 2.28.4, Python 3.13.11)
Hello from the pygame community.
```

* Reopening MiMo with `mimo -c` displays a blank window.
* Session data appears to remain intact.

## Observations

Processes still running:

```powershell
Get-Process mimo*
```

```
ProcessName: mimo
Working Set: ~493 MB
CPU Time: ~830s
```

```powershell
Get-Process python*
```

```
ProcessName: python
Working Set: ~240 MB
```

Session directory still exists:

```text
C:\Users\<user>\.local\share\mimocode\memory\sessions\ses_13c6b9671ffetHvYfzOngtr3dT
```

Contents:

```text
checkpoint.md
notes.md
tasks\
```

Files continue to exist after the UI disappears.

Example timestamps:

```text
checkpoint.md  9:47 PM
notes.md      10:21 PM
```

This suggests the session and memory subsystem may still be active even though the UI cannot be recovered.

## Additional Notes

This does not appear to be simple session loss.

Observed state:

* MiMo process alive
* Python process alive
* Session memory present
* Checkpoint files present
* UI blank / unrecoverable

Possible areas to investigate:

* Background process launching on Windows
* Terminal ownership restoration after interactive commands
* UI rehydration after reconnect (`mimo -c`)
* Pygame / long-running GUI applications
* Session recovery when child processes remain active

## Reproduction

1. Start MiMo Code.
2. Open a Python pygame project.
3. Ask MiMo to run the game.
4. MiMo reports:

```
[Interactive] Run <game> in background
```

5. Observe dashboard disappearance.
6. Run:

```powershell
mimo -c
```

7. Observe blank MiMo window instead of restored session.

**Update:**

Additional testing identified the trigger.

After asking MiMo to run a pygame application, the MiMo dashboard disappeared and `mimo -c` showed a blank window.

The session was not lost.

The Python process launched by MiMo remained running. As soon as the Python process was terminated, the MiMo session immediately returned and became usable again.

This appears to be a process-handling/UI restoration issue related to long-running GUI applications. Session persistence, checkpoints, and memory were preserved throughout the incident.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

MrRealORG:
Fixes #544

## Problem

On Windows, child processes were spawned with `detached: false`, which shares the parent's console. When a child process opens a GUI window (e.g., pygame) or writes to the console, it takes over the terminal, causing the MiMo dashboard (Ink/React TUI) to go blank. The session remains alive but the UI enters an unrecoverable state.

## Root Cause

`bash.ts` `cmd()`:
```diff
- detached: process.platform !== "win32",  // Windows: false
+ detached: true,                        // All platforms: true
+ windowsHide: process.platform === "win32",
```

`cross-spawn-spawner.ts` fallback:
```diff
- detached: command.options.detached ?? process.platform !== "win32",
+ detached: command.options.detached ?? true,
```

With `detached: false` on Windows, the child runs in the parent's console. A pygame window (or any GUI app) then takes over that console, blanking the TUI.

## Fix

- **bash.ts**: Use `detached: true` on all platforms. On Windows, add `windowsHide: true` to prevent a visible console window.
- **cross-spawn-spawner.ts**: Change the default `detached` fallback to `true`.

`detached: true` creates a new process group, isolating the child from the parent's console. Stdio is still piped (not shared), so output collection works normally. `windowsHide: true` prevents a visible console window on Windows.

The existing kill/cleanup logic already handles detached processes correctly (`taskkill /T /F` on Windows, `process.kill(-pid)` on Unix).

## Scope

Only 2 files, 5 insertions, 3 deletions. No behavior change on Unix (was already `detached: true`).


fengjikui:
I tested this patch locally because it touches a broad process-spawning path.

Applied the PR diff on top of current `origin/main`, then ran the focused package check:

```sh
PATH="$HOME/.bun/bin:$PATH" bun run --cwd packages/opencode typecheck
```

It fails in the two changed `ChildProcess.make(...)` calls in `src/tool/bash.ts`:

```text
src/tool/bash.ts(311,30): error TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type 'string' is not assignable to parameter of type 'TemplateStringsArray'.
src/tool/bash.ts(320,28): error TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type 'string' is not assignable to parameter of type 'TemplateStringsArray'.
```

After reversing the patch, the same package typecheck passes.

I think the root cause is that `windowsHide` is a Node `spawn` option, but `ChildProcess.make(...)` takes Effect's `ChildProcess.CommandOptions`, which currently exposes `detached` but not `windowsHide`. The spawner already maps Effect options to Node options and sets `windowsHide: process.platform === "win32"` in `cross-spawn-spawner.ts`.

So the minimal version may be: keep `windowsHide` only in the spawner layer, and change only the `detached` behavior where needed. It would also be good to add a regression check around `bash` timeout/abort/output collection, because this changes foreground Windows command behavior as well as the background pygame case.

fengjikui:
Thanks for the follow-up commit. I rechecked the latest head `18af4c1a` locally.

The previous `windowsHide` / `ChildProcess.make(...)` type error is fixed now:

```bash
cd packages/opencode
bun typecheck
bun test test/tool/bash.test.ts --timeout 60000
```

Results:

- package typecheck: pass
- `test/tool/bash.test.ts`: 21 pass, 0 fail
- `git diff --check`: clean

One remaining file-scope issue: `bun.lock` is still changed in this PR, with a very large unrelated diff:

```text
bun.lock | 768 +--------------------
```

The functional fix appears to be only:

```text
packages/opencode/src/effect/cross-spawn-spawner.ts
packages/opencode/src/tool/bash.ts
```

Unless the lockfile change is intentional for this Windows process fix, I would remove/revert `bun.lock` from the branch so maintainers can review just the process behavior change.

MrRealORG:
Good catch! The issue was `windowsHide` — it is not a valid property on `CommandOptions` in effect's `ChildProcess.make()`. It caused TypeScript to fall through to the template-literal overload, producing the confusing `TemplateStringsArray` error.

Removed `windowsHide` from both `cmd()` calls in `bash.ts`. The `windowsHide` option is already correctly applied at the `cross-spawn-spawner.ts` level where native Node.js spawn options are passed directly.

Typecheck now passes clean (12/12 tasks).

PR #591, opened 2026-06-14
[codex] fix GNU screen terminal input

### LINKED ISSUE #526: I use screen , then in screen when to open mimo, I cannot type !

fengjikui:
## Summary

- Detect GNU screen sessions as plain terminals in the TUI terminal compatibility helper.
- Preserve explicit `MIMOCODE_TUI_PLAIN=false` opt-out behavior.
- Add focused tests for `TERM=screen*`, `STY`, explicit overrides, and regular xterm behavior.

Fixes #526

## Why

The reporter confirmed that running MiMo with `MIMOCODE_TUI_PLAIN=1 mimo` fixes input inside GNU screen on WSL2 Ubuntu 24.04 with GNU screen 4.09.01. Current `main` only auto-enables the plain renderer for macOS Apple Terminal or explicit env override, so GNU screen can still use the advanced keyboard path that screen does not pass through cleanly.

There is an overlapping bundled PR (#586), but this PR intentionally keeps the scope to the confirmed #526 terminal input issue only. It does not include the unrelated status-bar or `.docx` reader changes from that bundle.

## Validation

From `packages/opencode`:

```bash
bun test test/cli/tui/terminal.test.ts --timeout 30000
bun typecheck
```

From repo root:

```bash
git diff --check origin/main...HEAD
```

Additional smoke check inside a real GNU screen session:

```text
{"TERM":"screen","STY":"89729.mimo-plain-check","plain":true}
```

Pre-push hook also ran root `bun turbo typecheck` successfully.


qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1008

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1053

PR #597, opened 2026-06-14
fix: properly wire i18n for mimo-auto model label + multiedit regression test

MrRealORG:
Supersedes #590 — addresses reviewer feedback from fengjikui.

Fixes #565

Changes vs #590:
- **local.tsx**: Now calls `t('tui.model.mimo_auto_free_promo')` via `useLanguage()` so zh/zht users see the localized model name instead of a hardcoded string
- **multiedit.test.ts**: Focused regression test for empty edits array (guard against `Cannot read properties of undefined`)
- **mimo.ts**: Hardcoded Chinese login label/instructions replaced with English
- **zht.ts**: 30 missing login/cli-provider i18n keys in Traditional Chinese
- **en.ts/zh.ts**: Added `tui.model.mimo_auto_free_promo` key

PR #599, opened 2026-06-14
fix(i18n): add 30 missing login/provider keys to es, ja, fr, ru

MrRealORG:
Follow-up to #597 — extends i18n coverage to all remaining TUI locales.

Previously login dialog, login flow, CLI provider, and model display keys
only existed in en.ts, zh.ts, and zht.ts. Spanish, Japanese, French, and
Russian users saw raw English fallbacks for the entire login/provider UI.

Adds 31 keys (30 login/provider + 1 model display) to each of:
- es.ts (Spanish)
- ja.ts (Japanese)
- fr.ts (French)
- ru.ts (Russian)


fengjikui:
I applied this patch on current `origin/main` for a quick compile check. The scope is nice and small, but there is one blocking issue: `fr.ts`, `ja.ts`, and `ru.ts` already contain several of these login/provider keys, and this PR appends duplicate object keys.

The package typecheck reports `TS1117: An object literal cannot have multiple properties with the same name` in those three files. Examples after applying the patch:

```text
fr.ts:411  "tui.dialog.login.mimo_free"
fr.ts:445  "tui.dialog.login.mimo_free"
fr.ts:417  "cli.providers.select"
fr.ts:465  "cli.providers.select"

ja.ts:363  "tui.dialog.login.mimo_free"
ja.ts:397  "tui.dialog.login.mimo_free"
ja.ts:369  "cli.providers.select"
ja.ts:417  "cli.providers.select"

ru.ts:423  "tui.dialog.login.mimo_free"
ru.ts:457  "tui.dialog.login.mimo_free"
ru.ts:429  "cli.providers.select"
ru.ts:477  "cli.providers.select"
```

`es.ts` looks like it is adding genuinely missing keys, but for `fr/ja/ru` the safer fix is probably to update the existing entries in place and add only the keys that are still missing, rather than appending the whole block. After that, `cd packages/opencode && bun typecheck` should be the useful gate for this PR.

MrRealORG:
Good catch! Removed all 15 duplicate keys from the appended blocks in `fr.ts`, `ja.ts`, and `ru.ts` (13 login/cli-provider keys that already existed). Kept the 16 genuinely new keys (login flow, import claude, model promo). `es.ts` was untouched.

Should compile clean now — no more TS1117 errors.

PR #601, opened 2026-06-14
fix: reduce default context window, improve plan mode triggering (#534, #531)

### LINKED ISSUE #531: Plan模式即使是一个简答的问答，也生成plan文件并询问是否需要执行
### Description

如图：

Mimo Code：
Image

OpenCode：

Image

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

WSL2 Fedora-43

### Terminal

WindowsTerminal

### LINKED ISSUE #534: Local Models Do Not Have Context Limits Defined, Causing Incorrect Context Usage Display and Potential Overflow

MrRealORG:
## Summary

Two targeted fixes addressing reported bugs:

### Fix #534: Local Models Do Not Have Context Limits Defined

**Problem:** `DEFAULT_CONTEXT_WINDOW` was set to `1,000,000` tokens. When a local/custom model (e.g. via llama.cpp) has no explicit `limit.context` in `mimocode.json` and is not found in the models.dev database, MiMo silently defaults to 1M tokens. This causes the context usage indicator to show wildly incorrect percentages — a user with a 180K context model at 22K tokens saw "2% used" when it was actually ~12%.

**Fix:** Changed `DEFAULT_CONTEXT_WINDOW` from `1,000,000` to `128,000`. This is a much safer default that:
- Matches the common context window size of many popular models
- Prevents the "silent overflow" where the agent keeps sending requests past the real limit
- Still warns users to configure explicit limits (existing warning log preserved)

### Fix #531: Plan mode triggers on simple Q&A

**Problem:** The `plan_enter` tool description said "Do NOT call this tool for simple, straightforward tasks" but the LLM still suggested plan mode for basic questions and explanations.

**Fix:** Strengthened `plan-enter.txt` with explicit additional guard clauses:
- "For questions, explanations, or conversations that do not require code changes"
- "When the user is only asking for information, clarification, or a simple answer"
- "For single-file edits, typo fixes, or trivial changes"
- Added IMPORTANT block: "Plan mode is ONLY for multi-step implementation tasks"

## Files Changed
- `packages/opencode/src/provider/provider.ts` — line 34: constant value
- `packages/opencode/src/tool/plan-enter.txt` — expanded negative examples


fengjikui:
I applied this PR on current `origin/main` and ran the focused provider checks locally:

```bash
cd packages/opencode
bun test test/provider/provider.test.ts -t "model limit defaults" --timeout 30000
bun test test/provider/provider.test.ts -t "opencode and opencode-go providers" --timeout 30000
bun typecheck
```

Results:
- `bun typecheck` passes.
- `git diff --check` passes.
- `opencode and opencode-go providers...` passes.
- The existing F41 fallback regression still fails. On current main it already expects `200_000` but receives `1_000_000`; with this PR applied it expects `200_000` but receives `128_000`:

```text
Expected: 200000
Received: 128000
```

So the regression appears stale on main already, but this PR should still update/rename that test if `128_000` is the intended new fallback. That would make the #534 behavior explicit for reviewers instead of leaving the fallback change covered only by implementation.

One process note: #582 already includes the same #534/#531 changes and currently claims `Fixes #534 #531 #518`, so maintainers may want to choose between this narrower PR and the broader #582.

Issue #721, opened 2026-06-15
EEXIST: file already exists, mkdir '.git/info' on Windows with OneDrive-synced projects


Jameswanttolearn:
## Description

MiMoCode v0.1.1 crashes on startup when run inside a Git repository whose ``.git/info`` directory carries the Windows ``ReadOnly`` attribute (common with OneDrive-synced folders).

## Plugins

None (clean install)

## OpenCode version

0.1.1

## Steps to reproduce

1. On Windows, have OneDrive syncing your project folder (which sets ``ReadOnly`` on ``.git/info``)
2. Open a terminal in that Git repository
3. Run ``mimo`` or ``mimo run "hello"``

**Reproduces without OneDrive too** — manually setting the attribute triggers it:

```powershell
attrib +R ".git\info"
mimo
```

## Screenshot and/or share link

```
Error: EEXIST: file already exists, mkdir 'C:\Users\...\project\.git\info'
    at async EN (B:/~BUN/root/chunk-vb4jhgk4.js:15:9958)
```

Full log from ``~\.local\share\mimocode\log\2026-06-15T100523.log``:

```
INFO  service=server-proxy directory=C:\Users\...\retropad creating instance
INFO  service=project directory=C:\Users\...\retropad fromDirectory
ERROR service=server-proxy name=Error message=EEXIST: file already exists, mkdir 'C:\Users\...\retropad\.git\info' fatal
```

## Operating System

Windows 11

## Terminal

PowerShell 5.1

## Root cause

``setupProjectIdEnvironment`` in ``packages/opencode/src/project/project.ts`` calls:

```ts
await nodeFs.mkdir(nodePath.dirname(excludeFile), { recursive: true })
```

where ``excludeFile`` is ``.git/info/exclude``. This should be a no-op when ``.git/info`` already exists, but on Windows ``mkdir`` with ``{ recursive: true }`` throws ``EEXIST`` when the target directory has the ``ReadOnly`` NTFS attribute. This is a known libuv/Node.js issue ([libuv#2897](https://github.com/libuv/libuv/issues/2897), [nodejs#43994](https://github.com/nodejs/node/issues/43994)). OneDrive sets this attribute on synced directories.

The same class of bug has been reported extensively in other tools:
- [anthropics/claude-code#33118](https://github.com/anthropics/claude-code/issues/33118) — mkdirSync without {recursive: true} on Windows
[...]

Issue #694, opened 2026-06-15
Session summary includes raw thinking blocks instead of final output only


HaleTom:
### Description

Session summary includes raw thinking/reasoning blocks in the output. It should only include the final output derived from that thinking, not the thinking process itself.

**Model:** Deepseek v4 Pro from Crof.AI

### Steps to reproduce

1. Start a MiMo Code session using Deepseek v4 Pro (Crof.AI provider)
2. Send a prompt that triggers model reasoning/thinking
3. Check the session summary — it contains raw thinking blocks instead of only the final response

### Screenshot 

<img width="576" height="335" alt="Image" src="https://github.com/user-attachments/assets/ed8df99f-b5f9-4cb5-8d6c-3c45dc539f32" />

### Expected behavior

Session summary should contain only the final output from the model's thinking, not the thinking/reasoning blocks themselves.

PR #730, opened 2026-06-15
fix: strip reasoning from compaction input (#694) and handle Windows mkdir EEXIST (#721)

MrRealORG:
## Summary

Fixes two bugs:

### #694 — Session summary includes raw thinking blocks

When the compaction LLM generates a session summary, it was receiving **reasoning/thinking parts** alongside the final output text. This caused the summary to contain raw internal thinking instead of only the final response.

**Root cause:** `compaction.ts` calls `MessageV2.toModelMessagesEffect()` which includes `type: "reasoning"` parts in the UIMessage array sent to the model.

**Fix:** Strip reasoning parts from the cloned messages before passing them to the compaction LLM. The compaction only needs the final output to produce an accurate summary.

### #721 — EEXIST crash on Windows with OneDrive-synced projects

On Windows, `fs.mkdir()` with `{ recursive: true }` throws `EEXIST` when the target directory (`.git/info`) has the `ReadOnly` NTFS attribute, which is commonly set by OneDrive on synced folders. This caused MiMo Code to crash on startup.

**Root cause:** Known libuv/Node.js issue ([libuv#2897](https://github.com/libuv/libuv/issues/2897), [nodejs#43994](https://github.com/nodejs/node/issues/43994)).

**Fix:** Wrap the `mkdir` call in a try-catch that ignores `EEXIST` errors, since the directory already existing is the desired state.

## Test plan

- [ ] Verify session compaction/summary no longer includes thinking block content
- [ ] On Windows with OneDrive: verify `mimo` starts without EEXIST crash in a synced repo
- [ ] Verify `.git/info/exclude` still gets `.mimocode-project-id` appended correctly

Note: The `@mimo-ai/cli` typecheck errors (`Duplicate identifier` in `plugin/index.ts`, missing `quickjs-emscripten-core` module) are **pre-existing** on `main` and unrelated to this change.


HaleTom:
I'm curious as to why you chose to include both of these seemingly unrelated fixes in the one PR.

It means that both get delayed if there's an issue with only one.

HaleTom:
How is this different to #700?

I'm curious as to the amount of HITL you're doing on your issues.

With free tokens from MiMo, it's literally free to say "hey agent, fix all the issues on opencode". 

I note there are check-boxes unticked here but also the PR is not marked as draft.  Intentional?

MrRealORG:
no <:

HaleTom:
@MrRealORG I assume that's the answer for the last question you parsed.  

Would you please review the others?

MrRealORG:
aa no <:

HaleTom:
@MrRealORG honestly, this doesn't make you look like a community player.  

Have an agent spam out hundeds of PRs but not engage when asked about them?

You want them merged?  "aa no".

PR #806, opened 2026-06-16
fix: clear writerFailures counter in resetThresholds

### LINKED ISSUE #734: SessionPrune.resetThresholds leaves stale checkpoint writer failure counters
### Problem

`SessionPrune` keeps several pieces of per-session checkpoint bookkeeping: `crossed`, `maxCrossed`, and `writerFailures`.

`resetThresholds(sessionID)` currently clears `crossed` and `maxCrossed`, but it leaves `writerFailures` untouched. Since `writerFailures` is keyed by the same `sessionID`, the same session can enter a fresh checkpoint threshold cycle while still carrying the previous consecutive writer failure count.

If the previous count is already close to `max_writer_failures`, a later checkpoint writer failure in that same session can reach the retry cap earlier than expected. It also leaves a small stale in-memory entry until a successful writer clears it or the process restarts.

### Expected behavior

When `resetThresholds(sessionID)` starts a fresh checkpoint threshold cycle for a session, all related per-session checkpoint threshold state should be reset together.

### Suggested fix

Clear the writer failure counter together with the other reset state:

```ts
writerFailures.delete(sessionID)
```

### Suggested test coverage

Add a targeted `SessionPrune.fireCheckpoints` test that:

1. records two checkpoint writer failures for a session,
2. calls `resetThresholds(sessionID)`,
3. verifies the next checkpoint threshold cycle starts with a fresh failure count instead of inheriting the previous value.

This keeps the fix scoped to the same session and avoids changing retry behavior across different sessions.

0xSkybreaker:
@
## Summary

Fixes #734 — `SessionPrune.resetThresholds(sessionID)` clears `crossed` and `maxCrossed` but leaves `writerFailures` untouched, causing stale failure counts to carry over into a fresh checkpoint threshold cycle after discard+rebuild.

## Changes

- **`packages/opencode/src/session/prune.ts`**: Add `writerFailures.delete(sessionID)` to `resetThresholds` so all per-session checkpoint threshold state resets together
- **`packages/opencode/test/session/prune.test.ts`**: Add targeted test that records 2 writer failures, calls `resetThresholds`, then verifies the next cycle has a full 3-retry budget (not just 1 remaining)

## Verification

The new test exercises the exact scenario described in the issue:
1. Record 2 checkpoint writer failures for a session (counter = 2)
2. Call `resetThresholds(sessionID)`  
3. Fire 3 more times — all 3 land (counter was reset to 0, not inherited from 2)
4. Fire once more — rejected at cap (counter = 3)

Without the fix, step 3 would only allow 1 more fire before hitting the cap.

🤖 Generated with [Claude Code](https://claude.com/claude-code)@


qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1017

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1044

PR #875, opened 2026-06-16
fix: replicate launcher's AVX2/musl detection in postinstall.mjs

### LINKED ISSUE #871: postinstall.mjs picks AVX2 binary on non-AVX2 x64 CPUs → SIGILL on first run (cache shadows launcher's runtime detection)
## Summary

  `@mimo-ai/cli@0.1.1` crashes with `SIGILL` (illegal instruction) on the very first invocation on x64 Linux CPUs that lack AVX2 (e.g. Intel Sandy Bridge / Ivy Bridge, AMD pre-Excavator). The `bin/mimo`
  launcher *does* implement runtime AVX2 detection and would correctly pick the `-baseline` variant, but `postinstall.mjs` writes a cache file at `bin/.mimocode` that points to the non-baseline build,
  and the launcher short-circuits to the cache before its detection runs.

  ## Reproduction

  - OS: Ubuntu 20.04.6 LTS, glibc 2.31, kernel 5.15
  - CPU: `Intel(R) Core(TM) i5-2400S` — Sandy Bridge, has AVX, **no AVX2 / BMI / FMA**
  - Node: v24.16.0, npm 11.13.0

  ```bash
  npm install -g @mimo-ai/cli
  mimo --version
  # (no output, exit 132 — SIGILL, core dumped)
  ```

  `strace` confirms `SIGILL {si_code=ILL_ILLOPN}` from the cached `.mimocode` binary.

  ## Root cause

  Two pieces of code disagree about which binary to use:

  **`bin/mimo` (launcher)** — does CPU feature detection:

  ```js
  // abridged
  const avx2 = supportsAvx2()              // reads /proc/cpuinfo on linux
  const baseline = arch === "x64" && !avx2
  // builds candidate names list with -baseline first when !avx2
  // ...but only AFTER:
  const cached = path.join(scriptDir, ".mimocode")
  if (fs.existsSync(cached)) {
    run(cached)                            // ← short-circuit; detection never runs
  }
  ```

  **`postinstall.mjs`** — does NOT do CPU feature detection, only `os.platform()` + `os.arch()`:

  ```js
  function detectPlatformAndArch() {
    // returns { platform: 'linux', arch: 'x64' }
  }
    // ... resolves and hardlinks to bin/.mimocode
  }
  ```

  So on a non-AVX2 x64 box, postinstall hardlinks `bin/.mimocode` → `mimocode-linux-x64/bin/mimo` (AVX2 build), and the launcher then runs that cached binary unconditionally → `SIGILL`.

  It also doesn't honour the `-musl` and `-baseline-musl` variants on Alpine.

  ## Fix (suggested)

  Have `postinstall.mjs` reuse the same selection logic as `bin/mimo` (AVX2 detection + musl detection + the same priority list), or simply skip writing the cache and always let the launcher pick. A
  minimal patch:

  ```js
  // In postinstall.mjs, replicate the launcher's `names` candidate list
  // and try them in order until one resolves; only then create the cache.
  ```

  Alternatively, just delete the postinstall caching step entirely — the launcher's resolution path (`findBinary` walking `node_modules`) already works, and avoiding the cache eliminates a class of
  stale/wrong-variant bugs (also helpful when the user's CPU changes, e.g. moving an npm-global between machines).

  ## Workaround for affected users

  ```bash
  CLI=$(npm root -g)/@mimo-ai/cli
  ln -f "$CLI/node_modules/@mimo-ai/mimocode-linux-x64-baseline/bin/mimo" "$CLI/bin/.mimocode"
  ```

  Or set `MIMOCODE_BIN_PATH` to the baseline binary explicitly (the launcher honours this env var first, before the cache).

  ## Impact

  Anyone on x64 hardware older than ~Haswell (2013) — Sandy Bridge, Ivy Bridge, early AMD Bulldozer/Piledriver — will hit this. Symptom is silent (no error message, just exit 132 / core dump), so it'll
  mostly produce confused issue reports rather than clean stack traces.

  ---

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

GautamKumarOffical:
Fixes #871

`postinstall.mjs` always hardcodes the non-baseline, non-musl binary package name. On non-AVX2 x64 CPUs (Sandy Bridge, Ivy Bridge, pre-Excavator AMD) this writes a cache file pointing to the wrong binary, and the launcher short-circuits to that cache before its own CPU detection runs, causing SIGILL.

**What changed:**

Replicated the launcher's (`bin/mimo`) binary selection logic in `postinstall.mjs`:
- Added `supportsAvx2()` â€” checks `/proc/cpuinfo` on Linux, `sysctl` on macOS
- Added `detectMusl()` â€” checks for Alpine or musl-linked ldd
- `findBinary()` now builds the same candidate list with correct priority ordering (baseline first for non-AVX2, musl variants for Alpine) and picks the first that resolves

The cached `bin/.mimocode` now points to the correct binary for the user's CPU.


qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1021

qiaozongming:
https://github.com/XiaomiMiMo/MiMo-Code/pull/1040

Issue #494, opened 2026-06-13
[bug]: mimo stats includes imported Claude Code sessions in all metrics


llupRisinglll:
### Description

`mimo stats` counts imported Claude Code sessions in all metrics (sessions, days, tokens, cost, tool usage). Users who just installed MiMo see inflated stats from their old Claude Code history.

### Steps to reproduce

1. Install MiMo on a machine that has Claude Code sessions in `~/.claude/projects`
2. Run `mimo stats`
3. Sessions, days, tokens, and cost are inflated by the imported Claude Code data

### Screenshot and/or share link

Sessions count shows 216 (204 imported + 12 native), Days shows 32 (from oldest imported session).

### Operating System

Linux

PR #492, opened 2026-06-13
fix(cli): exclude imported sessions from stats by default

llupRisinglll:
### Issue for this PR

Closes #494

### Type of change

- [x] Bug fix

### What does this PR do?

`mimo stats` was counting imported Claude Code sessions in all metrics. This change filters them out by default using `WHERE id NOT IN (SELECT session_id FROM claude_import)` -- a subquery against the import tracking table rather than version-string matching, so any future importer is automatically excluded.

Added `--all` flag to include imported sessions when needed.

Related to #377 (stop auto-import on startup) -- that PR removes the auto-import trigger so new sessions aren't pulled in on every CLI invocation. This PR handles the stats side: even if users manually run `import-claude`, stats stays clean by default. The two are complementary and independent.

### How did you verify your code works?

Built and ran from source against a DB with 204 imported + 13 native sessions:

- `mimo stats` -> Sessions: 13 (native only), Days: 1
- `mimo stats --all` -> Sessions: 217 (all), Days: 32

Unit tests:
- `bun test test/cli/stats-exclude-imported.test.ts` -- 1 pass
- `bun test test/cli/import.test.ts` -- 5 pass (existing tests unaffected)

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #1000, opened 2026-06-18
fix(cli): exclude imported sessions from stats by default

qiaozongming:
Resubmitting community contribution; the original PR (#492) was auto-closed by a branch history rewrite.

### DIFF
--- a/packages/opencode/src/cli/cmd/stats.ts
+++ b/packages/opencode/src/cli/cmd/stats.ts
@@ -2,7 +2,7 @@ import type { Argv } from "yargs"
 import { cmd } from "./cmd"
 import { Session } from "../../session"
 import { bootstrap } from "../bootstrap"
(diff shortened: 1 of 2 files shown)


llupRisinglll:
Closes #494

llupRisinglll:
师兄 @qiaozongming, the stats query points at `claude_import`, which was renamed to `external_import` in [2ecfe94](https://github.com/XiaomiMiMo/MiMo-Code/commit/2ecfe943926c5e340a1d1cd6a2afb048caf39aac) with the old table dropped, so the exclusion is currently a no-op and should read `external_import WHERE source = 'cc'`. The test also needs `MIMOCODE_DISABLE_CLAUDE_IMPORT=true` in the env because startup auto-imports first; want me to open a fresh PR with these fixes?

PR #373, opened 2026-06-12
Enable TUI mouse scroll in plain terminals

YOMXXX:
## Summary
- keep Apple Terminal/plain rendering mode, but no longer disables mouse scroll by default
- preserve explicit opt-outs via tui config mouse=false or MIMOCODE_DISABLE_MOUSE
- add regression coverage for plain terminal mouse enablement

Closes #369.
Closes #381.

## Verification
- bun test test/cli/tui/terminal.test.ts test/cli/tui/thread.test.ts --timeout 30000
- bun typecheck
- git diff --check

Note: pushed with --no-verify because the repository pre-push hook runs root-level turbo typecheck, which is blocked locally by the missing optional turbo binary; package-level verification above passed.

PR #1076, opened 2026-06-18
Enable TUI mouse scroll in plain terminals

qiaozongming:
Resubmitting community contribution; the original PR (#373) was auto-closed by a branch history rewrite.


manoslinh:
Tested on macOS Apple Terminal/plain terminal mode. Default config now scrolls MiMo Code TUI history as expected. `tui.mouse=false` disables TUI mouse capture, so wheel input falls back to Apple Terminal native scrollback instead of MiMo history.

PR #160, opened 2026-06-11
fix(tui): preserve command keybinds while prompt suggestion is visible

ICOM725:
Fixes #124
Refs #65

When a ghost prompt suggestion is visible, the prompt suspends command keybinds
so Tab can reach the textarea and accept the suggestion. The previous suspend
state was too broad for ghost suggestions — it also blocked unrelated shortcuts
like Ctrl+P (command palette), F-keys, and leader sequences.

Keep the existing full suspension mode for autocomplete, where Ctrl+P/Ctrl+N are
handled by the autocomplete list itself. Add a narrower input suspension mode for
ghost suggestions so only the prompt-owned keys (Tab to accept, Escape to dismiss)
fall through while the command palette shortcut keeps working.

Verification:
- bun run --cwd packages/opencode typecheck
- bun test test/cli/tui/keybind-plugin.test.ts (5 tests pass)

PR #1107, opened 2026-06-18
fix(tui): preserve command keybinds while prompt suggestion is visible

qiaozongming:
Resubmitting community contribution; the original PR (#160) was auto-closed by a branch history rewrite.


ICOM725:
Thanks for carrying this forward. Since this PR is based on the fix from #160, could you please keep attribution to my original PR/commit, for example with a Co-authored-by trailer or by preserving the original author?

PR #1196, opened 2026-06-22
fix(test): fix CI test failures on main

onlyfeng:
## Summary
- Fix provider.test.ts: MimoFreeAuthPlugin only exists in internal overlay, skip opencode disable assertions when mimo provider absent
- Fix whitelist.test.ts: filterCompactedEffect defaults to main slice, subagent messages live in their actorID slice; use agentID='*'

### DIFF
--- a/packages/opencode/test/provider/provider.test.ts
+++ b/packages/opencode/test/provider/provider.test.ts
@@ -2638,12 +2638,13 @@ test("opencode and opencode-go providers are disabled by MimoFreeAuthPlugin", as
 
   // MimoFreeAuthPlugin always pushes opencode/opencode-go into disabled_providers,
   // so they should not appear even when the user supplies an apiKey or auth record.
-  expect(opencodeProviderPresent(providers)).toBe(false)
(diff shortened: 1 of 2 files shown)


yanyihan-xiaomi (approved):
**Approve**

This is a focused, test-only fix that correctly addresses the two CI failures on `main`. Both changes target the open-source vs. internal-build divergence rather than papering over real failures.

- `provider.test.ts`: Correctly identifies that `MimoFreeAuthPlugin` lives in the private overlay (`src/ext/`) and is absent in the open-source build, so `opencode`/`opencode-go` stay loaded there. Gating the `disabled_providers` assertions behind `if (mimo)` is the right approach, and the updated comment accurately documents the reason.
- `whitelist.test.ts`: The root cause is well understood — `filterCompactedEffect` defaults to the `main` slice while subagent messages live in their own `actorID` slice. Passing `{ agentID: "*" }` to include the subagent slice is the correct fix, and the inline comments make the intent clear.

The diff is small (+11/−6), comments explain the *why*, and no production code is touched. No concerns — approving.

PR #1268, opened 2026-06-24
fix(session): preserve failed subtask metadata

onlyfeng:
## Summary

Preserve tool metadata when subtask execution fails so the TUI can keep navigation context for failed subtasks.

## Root Cause

Subtask tool parts already carry metadata such as `sessionId` and `model`, but pending-state error handling dropped metadata before writing the final error state. Running tool-call error handling also overwrote metadata when adding the recoverable marker.

## Changes

- Preserve existing metadata when a running tool call transitions to error.
- Merge `recoverable: true` into existing metadata instead of replacing it.
- Preserve subtask metadata even when the tool part fails from pending state.
- Update the regression test to run with an Actor layer so the actor tool path is exercised.

## Validation

- `bun typecheck`
- `bun test --timeout 60000 test/session/prompt-effect.test.ts -t "failed subtask preserves metadata"`

## CI Status Note

- `lint` and `typecheck` pass on this PR.
- Full `unit` currently fails, but `upstream/main@ee4ddd2` already has a failing `test` workflow (run `28071223407`).
- This PR fixes its target failure: `failed subtask preserves metadata on error tool state` is absent from this PR unit failure list.
- The remaining `Claude Code local MCP server is pending until explicitly connected` failure is covered by companion PR #1269, not this metadata change.
- The `CheckpointContext producer`, `parentSessionID end-to-end`, and `session.llm.stream` failures are also present on current `upstream/main` and are outside this diff.
- One additional `WorkflowRuntime cancel cascade` timeout appeared in this PR run; this diff does not touch workflow runtime cancellation, so it should be tracked separately if it reproduces.

### DIFF
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -286,17 +286,20 @@ export const layer: Layer.Layer<
       const failToolCall = Effect.fn("SessionProcessor.failToolCall")(function* (toolCallID: string, error: unknown) {
         const match = yield* readToolCall(toolCallID)
         if (!match || match.part.state.status !== "running") return false
-        // Agent-recoverable failures (bad args, malformed call, unknown task/actor
-        // id) carry a marker the TUI reads to render them muted instead of as a red
-        // error block. The full actionable message still flows to the model.
         const recoverable = isRecoverableError(error)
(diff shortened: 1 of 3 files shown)


onlyfeng:
Closing: upstream main already preserves failed subtask metadata via `getToolResultMetadata()` in processor.ts failToolCall. The prompt.ts metadata handling is functionally equivalent. No remaining action needed.

PR #1296, opened 2026-06-24
fix: green the failing unit test suite

lLukDreams:
### Issue for this PR

Closes #   

### Type of change

- [x] Bug fix

### What does this PR do?

Repairs the unit suite. The failures were a mix:

**Stale tests** (code changed, assertion didn't):
- Drop `/root` from the forbidden-dirs assertion — it was intentionally removed from the blocklist.
- Update the `cache_control` expectation to the rolling-tail marker placement the provider transform now uses.

**Real bugs:**
- A subtask that fails before `onReady` lost its metadata; seed `sessionId`/`model` up front so the error tool state keeps it.
- `.claude.json` MCP servers were eagerly connected instead of lazy `pending` — the `mcp_origins` "claude" marker wasn't honored.
- The checkpoint writer read the Actor service only via the global `spawnRef`, which goes stale when a memoized Actor layer is torn down by instance disposal in an earlier test. Resolve from the live context via `Effect.serviceOption` (optional, so no layer cycle).
- Workflow cancel could spawn orphans and hang with a wide fan-out: the promise-based agent semaphore kept starting queued agents as reclaim freed permits, and `Actor.cancel` awaited a fiber unwind that never finishes when stuck on a hung LLM stream. Bail queued spawns once cancelling, and bound the interrupt.

**Test isolation:**
- History backfill inspects all FTS rows but only cleared the shared DB in `afterEach`; rows from earlier shard files leaked in. Clear in `beforeEach` too.

Ran the full suite on Linux across all 4 shards (matching the sharded `ubuntu-latest` CI) — green, 0 failures. The workflow cancel fix specifically: 20/20 with the concurrency ceiling pinned below the fan-out to force queuing. `typecheck` clean.

### Checklist

### DIFF
--- a/packages/opencode/src/actor/spawn.ts
+++ b/packages/opencode/src/actor/spawn.ts
@@ -702,7 +702,14 @@ export const layer = Layer.effect(
           concurrency: "unbounded",
           discard: true,
         })
-        yield* state.cancelActor(sessionID, actorID)
+        // Bound the work-fiber interrupt. cancelActor interrupts the actor's runner
+        // and awaits its full unwind; a fiber stuck unwinding (e.g. blocked on a
+        // hung LLM response stream) would otherwise hang the entire cancel sweep
+        // indefinitely. The interrupt signal is delivered regardless — we just stop
+        // awaiting completion after a grace period and let it finish detached. The
+        // registry status below still records the cancellation, which is the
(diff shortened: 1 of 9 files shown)


lLukDreams:
@yanyihan-xiaomi

PR #1324, opened 2026-06-25
fix(agent): enforce plan mode write restrictions without mutating the tool schema

yanyihan-xiaomi:
## Summary

Plan mode now enforces its read-only intent at the permission layer **without ever removing a tool from the model's schema**: writes are blocked at call time (edit to non-plan files denied; `bash`/`change_directory`/`workflow` require confirmation), while every tool stays present so entering plan mode does not mutate the schema and invalidate the prefix cache. Plan can still spawn research subagents (e.g. `explore`), constrained to read-only so writes can't be delegated around the restriction. It also fixes the genuine root cause of the report: a persisted "always" approval (e.g. an edit approved in build mode) could out-rank plan's `edit:deny` and leak a write through.

> Plan 模式现在在权限层强制只读，但**绝不从工具 schema 移除任何工具**：写操作在调用时被拦，所有工具仍保留在列表中，因此切入 plan 不会让工具列表抖动、不会使 prefix cache 失效。Plan 仍可 spawn 只读的调研 subagent。同时修复了真正的根因：build 模式下持久化的 "always" 审批会越过 plan 的 `edit:deny` 泄漏一次写入。

## Relationship to #1301

This PR supersedes the third-party #1301 (which also targeted #870 / #1274). #1301 was directionally correct but had two real defects:

1. **Tool-list mutation / prefix-cache invalidation.** #1301 set `bash/task/workflow/change_directory` to a bare `"*":"deny"`. `Permission.disabled()` strips a tool only when its last-matching rule is exactly `pattern==="*" && action==="deny"`, so entering plan mode removed those tools from the list mid-session — the exact instability #1207 fixed for `plan_enter`/`plan_exit`. This PR uses `"ask"` (stays visible; prompts at call time; headless agents get `DeniedError`) or a deny-with-non-`"*"`-exception (`edit` keeps its plan-file allow).
2. **Plan can't research.** #1301's `task:"deny"` blocked spawning `explore`/`general`. This PR keeps `task` allowed and constrains spawned subagents to read-only via a new `subagentToolAllowlist` field.

Additionally #1301 used two `agent.name === "plan"` runtime checks; this PR encodes the restriction as data on the agent definition (`hardPermission` + `subagentToolAllowlist`), applied via a single `runtimePermission` helper with no name-checks. Recommend closing #1301.

> 本 PR 取代第三方 #1301（同样针对 #870 / #1274）。#1301 思路正确但有两处实现问题：裸 `"*":"deny"` 会让工具从 schema 消失，切模式时抖动并使 prefix cache 失效；`task:"deny"` 又让 plan 无法 spawn 调研 subagent。本 PR 改用 `ask` + 只读 subagent 白名单，并以数据驱动的 `hardPermission` 取代两处 `=== "plan"` 硬编码。建议关闭 #1301。

## Implementation

- `Agent.Info` gains `hardPermission` (non-overridable rules re-appended after the user/session merge so they always win) and `subagentToolAllowlist` (tool allowlist forced on spawned subagents); exports a `READONLY_TOOLS` constant.
- `runtimePermission(agent, session)` = `merge(agent.permission, session, agent.hardPermission)`, wired into all 5 evaluation sites (`llm.ts` ×2, `prompt.ts` ×2, `debug/agent.ts`).
- `permission/index.ts`: the `ask` loop evaluates the ruleset alone first — `deny` short-circuits, approvals can only upgrade `ask`→`allow`, never override `deny`.
- `actor.ts`: reads the spawning agent's `subagentToolAllowlist` and forces it onto the child before spawn.

> 实现：`Agent.Info` 新增 `hardPermission`（合并后再追加、永远胜出的不可覆盖规则）与 `subagentToolAllowlist`（强制施加给 subagent 的白名单），导出 `READONLY_TOOLS`；`runtimePermission` 接线到全部 5 个求值点；`permission/index.ts` 让 `deny` 短路、审批只能把 `ask` 升级为 `allow`；`actor.ts` 在 spawn 前强制施加父 agent 的只读白名单。

## Verification

- `bun test test/agent/agent.test.ts test/permission/next.test.ts test/tool/actor.test.ts` — 140 pass / 0 fail
- `bun typecheck` — clean
- **Prefix-cache stability guard**: enumerates the real registered tool universe and asserts every primary agent strips the same set as build via `Permission.disabled(runtimePermission(agent))`. Manually verified to fail when plan's `bash:"ask"` is regressed to `bash:"deny"` — so it catches any future bare deny that would mutate the tool list.
- **`READONLY_TOOLS` ⊆ registry guard**: asserts every entry resolves to a real registered tool. It caught entries absent from the default registry — `list` (never a tool id) and the gated `websearch`/`codesearch`/`lsp` — which were removed, leaving `read, glob, grep, webfetch, memory, history`.

> 验证：相关测试 140 全过、typecheck 干净；prefix-cache 稳定性守卫枚举真实工具全集，断言每个 primary agent 的剥离集与 build 相同，并已验证当 plan 的 `bash:"ask"` 退化为 `bash:"deny"` 时立即失败；新增的 `READONLY_TOOLS` ⊆ registry 守卫断言每一项都能在 registry 解析到，借此剔除了默认 registry 中不存在的 `list` 及受开关控制的 `websearch`/`codesearch`/`lsp`。

See `docs/compose/reports/plan-mode-write-restrictions.md` for the full report.

Co-authored-by: MiMo-Code <noreply@mimo.xiaomi.com>

### DIFF
--- a/docs/compose/plans/2026-06-25-plan-mode-write-restrictions.md
+++ b/docs/compose/plans/2026-06-25-plan-mode-write-restrictions.md
@@ -1,0 +1,438 @@
+# Plan Mode Write Restrictions — Hardened Implementation Plan (v2)
+
+> [!NOTE]
+> This document may not reflect the current implementation.
+> See the final report for up-to-date state:
+> [Final Report](../reports/plan-mode-write-restrictions.md)
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use compose:execute to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make plan mode's write restrictions un-bypassable WITHOUT ever removing a tool from the model's schema — deny actual writes at call time, keep every tool present so the tool list never mutates on mode switch (which would invalidate prefix cache and destroy context).
+
+**Architecture:** Add two declarative fields to `Agent.Info`: `hardPermission` (rules re-appended after the user/session merge so they always win) and `subagentToolAllowlist` (read-only toolset forced on any subagent the agent spawns). A single `runtimePermission(agent, sessionPermission)` helper replaces every `Permission.merge(agent.permission, …)` call site. Crucially, plan's hard rules use only `ask` and `deny-with-non-"*"-exception` forms — NEVER a bare `{"*":"deny"}` — because `Permission.disabled()` strips a tool from the schema ONLY when the last-matching rule is exactly `pattern==="*" && action==="deny"`. The persisted-approval bypass (an `always`-approved edit out-ranking plan's `edit:deny`) is closed in the permission `ask` loop (already committed as Task 0).
+
+**Tech Stack:** TypeScript, Effect, Bun test, Zod schema (`Agent.Info`), the existing `Permission` ruleset system.
+
+## Global Constraints
+
+- Run all commands from `packages/opencode` — tests are guarded against running from repo root.
+- Typecheck with `bun typecheck` from `packages/opencode`, never `tsc` directly.
+- **No tool may be removed from the schema in plan mode.** Forbidden: any plan rule of the form `<tool>: "deny"` or `<tool>: { "*": "deny" }` with no later non-`"*"` allow exception. Reference: PR #1207 made plan_enter/plan_exit "always visible" precisely because permission-based hiding caused tool-list mutation. We must not reintroduce that.
+- Verification that a tool stays in the schema = `Permission.disabled([tool], runtimePermission(plan, []))` returns an empty set for that tool.
+- `edit` keeps its existing shape `{ "*": "deny", "<plan-file globs>": "allow" }` — the non-`"*"` allow exception keeps the edit tool in the schema while denying writes to non-plan files at call time.
+- `bash`, `change_directory`, `workflow` → `"ask"` (stays in schema; interactive prompt at call time; headless agents get DeniedError via the existing `interactive:false` path).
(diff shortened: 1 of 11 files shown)


yanyihan-xiaomi:
Converted to **draft**. Per design discussion, opencode/mimocode favors trusting the model and minimizing user prompts — the permission layer is a **backstop**, not an active gate. This PR's gating of `bash`/`change_directory`/`workflow` via `ask` (and forcing read-only subagents) goes beyond that backstop role and can cause repeated confirmation prompts for read-only shell use.

A **simplified alternative** is being opened that keeps only the genuine fix: ensuring user/session config cannot bypass plan mode's `edit`/`write` restriction (plus the persisted-approval-vs-deny root-cause fix). It does **not** put `bash`/`change_directory`/`workflow` into `ask`/`deny`. Both will be left as drafts for human experts to compare.

> 已转 draft。按设计讨论：opencode/mimocode 更信任模型、尽量少打断用户，权限层是**兜底**而非主动管控。本 PR 把 `bash`/`change_directory`/`workflow` 用 `ask` 拦截（并强制 subagent 只读）超出了兜底定位，会对只读 shell 造成反复确认。即将开一个**简化版**：只保留真正的修复——防止用户/会话配置绕过 plan 的 `edit`/`write` 限制（外加持久化审批不越过 deny 的根因修复），**不**把 bash/change_directory/workflow 设为 ask/deny。两者都留作 draft 供人类专家对比。

PR #1388, opened 2026-06-26
fix(tui): handle session page keybinds

### LINKED ISSUE #1372: TUI messages_page_up / messages_page_down keybinds have no listener — page keys silently do nothing in session view
# Bug Report: TUI `messages_page_up` / `messages_page_down` keybinds are registered but have no listener — page keys silently do nothing in the session view

## Summary

The TUI keybinds `messages_page_up` (default: `pageup,ctrl+alt+b`) and
`messages_page_down` (default: `pagedown,ctrl+alt+f`) are properly registered
in `packages/opencode/src/config/keybinds.ts` (lines 40-41) and appear in
the command palette, but **no `useKeyboard` listener in the session view
ever calls `keybind.match()` for them**. Pressing PageUp / PageDown / the
alternative bindings in the session view does nothing.

The "intermittent" symptom: page keys appear to work when certain dialogs
(e.g. `dialog-select`) are open, because those dialogs have their own
`useKeyboard` block with hardcoded `evt.name === "pageup"` / `"pagedown"`
checks. The session view itself has no such handler.

## Reproduction

1. Start any session: `mimo` (or `mimo run --agent build ...`)
2. Generate enough conversation history to make the message pane scrollable
   (e.g. run a 20+ turn agent loop or load a session with 30+ messages)
3. Try to scroll the message view by pressing:
   - `PageDown` (or `ctrl+alt+f`)
   - `PageUp` (or `ctrl+alt+b`)
4. **Expected:** the message pane scrolls by one page
5. **Actual:** nothing happens
6. Open the command palette (`ctrl+k` typically) and look for "Page down"
   / "Page up" — they show up with their keybind labels, confirming the
   binding is registered. But the same keypress that should "just work"
   when not in command palette mode does nothing.

## Root cause

Two-layer system in the TUI:

1. **Registration layer** — `packages/opencode/src/config/keybinds.ts:40-41`:
   ```typescript
   messages_page_up: keybind("pageup,ctrl+alt+b", "Scroll messages up by one page"),
   messages_page_down: keybind("pagedown,ctrl+alt+f", "Scroll messages down by one page"),
   ```

2. **Listening layer** — each component needs its own `useKeyboard` block
   that calls `keybind.match("messages_page_down", evt)` and reacts.

The session view's `useKeyboard` block at
`packages/opencode/src/cli/cmd/tui/routes/session/index.tsx:298` only
handles `app_exit`:

```typescript
useKeyboard((evt) => {
  if (!session()?.parentID && currentAgentID() === "main") return
  if (keybind.match("app_exit", evt)) {
    // ... abort session logic ...
  }
})
```

There's no `keybind.match("messages_page_up", ...)` or
`keybind.match("messages_page_down", ...)` call. The scroll object is in
scope (`scroll.scrollBy(amount)` is already used at line 722 / 733 for the
half-page command palette actions) but no keypress routes to it.

The command palette registers "Page up" / "Page down" entries at line
716-729 of the same file, each pointing at the `messages_page_*` keybind
IDs. But these palette entries are just *display labels* — they don't wire
the keystroke. The wiring is missing.

## Why "intermittent" instead of "always broken"

`packages/opencode/src/cli/cmd/tui/ui/dialog-select.tsx:213-214` has:

```typescript
if (evt.name === "pageup") move(-10)
if (evt.name === "pagedown") move(10)
```

This is a *direct* key-name check, not a `keybind.match()` call. When a
dialog-select is open (e.g. for slash command autocomplete, file picker, or
model picker), the page keys work because the dialog receives them. When
the dialog is closed and the session view is focused, the page keys fall
through to the session view, which has no handler. That's the
"intermittent" symptom.

This is a systemic anti-pattern in the TUI codebase: a keybind can be
registered in `keybinds.ts` (and appear in command palette) **without any
component listening for it**. There's no compile-time or startup-time
check that all registered keybinds have at least one listener.

### A. Add the missing listener in the session view

In `packages/opencode/src/cli/cmd/tui/routes/session/index.tsx`, extend
the existing `useKeyboard` block at line 298 to handle the two page
keybinds:

```typescript
useKeyboard((evt) => {
  if (!session()?.parentID && currentAgentID() === "main") return
  if (keybind.match("app_exit", evt)) {
    // ... existing abort logic ...
  }
  if (keybind.match("messages_page_down", evt)) {
    scroll.scrollBy(scroll.height)
    evt.preventDefault()
  } else if (keybind.match("messages_page_up", evt)) {
    scroll.scrollBy(-scroll.height)
    evt.preventDefault()
  }
})
```

### B. Add a startup-time audit that flags keybinds with no listeners

Walk the registered `Keybinds` config at TUI startup. For each keybind ID,
check whether the keybind name (or one of its parsed key strings) appears
in any `useKeyboard` body via static analysis (regex against source files
under `packages/opencode/src/cli/cmd/tui/`). If a keybind has zero listener
matches, log a warning. This catches the same anti-pattern in the future
before users file bug reports like this one.

### C. Add a `Keybind.referenced()` helper

Expose a function like `keybind.isReferenced("messages_page_down")` that
returns true if any TUI source file contains a `keybind.match("messages_page_down", ...)`
call. Combine with (B) above to make the audit a single line.

## Workaround used locally (not requesting merge)

For the record, we patched locally at
`packages/opencode/src/cli/cmd/tui/routes/session/index.tsx` by adding the
two `keybind.match` checks to the existing `useKeyboard` block. The fix is
4 lines. After rebuild + deploy, pageup/pagedown work in the session
view as expected.

## Environment

- mimo version: 0.1.3 (binary built from local fix commit)
- OS: macOS 15.5 (M4 Pro, arm64)
- TUI: opentui solid
- Affected TUI routes: `routes/session/index.tsx` (the main session
  view). Other routes may have similar "registered but unlistened"
  keybinds but were not investigated.

## Severity

**Medium** — page keys are an expected shortcut for scrolling long
sessions. Users encountering this will assume the feature is missing from
the TUI rather than the binding being unwired, and will switch to mouse
scroll or arrow keys. Not blocking, but confusing and undermines trust
in the keybind system shown in command palette.

mturac:
## Summary
- handle `messages_page_up` and `messages_page_down` directly in the session view when no dialog is active
- align page and half-page scroll distances with their keybind labels
- add regression coverage for the scroll distance contract

Fixes #1372.

## Validation
- `bun test test/cli/tui/session-scroll.test.ts test/cli/tui/keybind-plugin.test.ts`
- `bun typecheck`
- pre-push hook: `bun turbo typecheck`

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -105,6 +105,23 @@ const GO_UPSELL_WINDOW = 86_400_000 // 24 hrs
 const QUEUE_TOKEN_PLAN_LAST_SEEN_AT = "queue_token_plan_last_seen_at"
 const QUEUE_TOKEN_PLAN_WINDOW = 86_400_000 // 24 hrs
 
+type MessageScrollKeybind =
+  | "messages_page_up"
+  | "messages_page_down"
+  | "messages_half_page_up"
+  | "messages_half_page_down"
+
+const MESSAGE_SCROLL_DELTA = {
+  messages_page_up: (height: number) => -height,
+  messages_page_down: (height: number) => height,
+  messages_half_page_up: (height: number) => -height / 2,
+  messages_half_page_down: (height: number) => height / 2,
+} satisfies Record<MessageScrollKeybind, (height: number) => number>
+
+export function messageScrollDelta(keybind: MessageScrollKeybind, height: number): number {
+  return MESSAGE_SCROLL_DELTA[keybind](height)
+}
+
 const context = createContext<{
   width: number
   sessionID: string
@@ -330,7 +347,21 @@ export function Session() {
   })
 
   useKeyboard((evt) => {
-    if (currentAgentID() === "main") return
+    if (dialog.stack.length > 0 || evt.defaultPrevented) return
+    if (keybind.match("messages_page_up", evt)) {
+      evt.preventDefault()
+      scroll.scrollBy(messageScrollDelta("messages_page_up", scroll.height))
+      return
+    }
+    if (keybind.match("messages_page_down", evt)) {
+      evt.preventDefault()
+      scroll.scrollBy(messageScrollDelta("messages_page_down", scroll.height))
+      return
+    }
+  })
+
+  useKeyboard((evt) => {
+    if (!session()?.parentID && currentAgentID() === "main") return
     if (keybind.match("app_exit", evt)) {
       const status = sync.data.session_status?.[route.sessionID]
       if (status && status.type !== "idle") {
@@ -787,7 +818,7 @@ export function Session() {
       category: "session",
       hidden: true,
       onSelect: (dialog) => {
-        scroll.scrollBy(-scroll.height / 2)
+        scroll.scrollBy(messageScrollDelta("messages_page_up", scroll.height))
         dialog.clear()
       },
     },
@@ -798,7 +829,7 @@ export function Session() {
       category: "session",
       hidden: true,
       onSelect: (dialog) => {
-        scroll.scrollBy(scroll.height / 2)
+        scroll.scrollBy(messageScrollDelta("messages_page_down", scroll.height))
         dialog.clear()
       },
     },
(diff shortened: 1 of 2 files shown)


mturac:
Closing this to reduce stale open PR surface and keep the queue focused on higher-impact implementation fixes. I can revisit the session keybind behavior later if maintainers signal interest in this area.

PR #1400, opened 2026-06-27
fix(test): settle fake checkpoint writer so main-slice test drains cleanly

onlyfeng:
## Summary

`checkpoint-main-slice` test: the recording actor's spawn outcome never resolved,
leaving `tryStartCheckpointWriter`'s settlement-watcher fiber suspended until scope
close force-interrupted it — a flaky source under `it.live`. Settle the outcome so
the watcher completes (`writers.delete`) and assert the writer drains to `no-writer`.

No production change — `actor/spawn.ts` is untouched; the test's spawn-ref finalizer
stays `= undefined`, consistent with the sibling checkpoint tests.

## Verification

- `bun typecheck` — exit 0
- `bun test test/session/checkpoint-main-slice.test.ts` — 1 pass / 13 assertions

### DIFF
--- a/packages/opencode/test/session/checkpoint-main-slice.test.ts
+++ b/packages/opencode/test/session/checkpoint-main-slice.test.ts
@@ -29,8 +29,9 @@ const ref = {
 
 // Capture the SpawnInput passed to actor.spawn so the test can inspect the
 // ForkContext (specifically watermarkMsgID) that tryStartCheckpointWriter
-// computed. The outcome never resolves — we only care about what was passed
-// in, not what the writer produces.
+// computed. The outcome resolves immediately so the checkpoint writer's
+// settlement watcher runs to completion (writers.delete) instead of leaving a
+// fiber suspended past scope close — the test asserts that drain at the end.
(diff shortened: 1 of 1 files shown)


onlyfeng:
Note for reviewers — the failing `unit (shard 4/4)` checks are pre-existing flakiness on the base, not introduced by this PR.

- The two failing tests (`CheckpointContext producer ... after settle` and `parentSessionID end-to-end ... splitover`) already fail on the base commit `0abfd8a` on `main` (push CI at 2026-06-27 02:30), before any change here.
- This PR's only change is `test/session/checkpoint-main-slice.test.ts`; `src/actor/spawn.ts` is untouched. Locally these all pass: `checkpoint-main-slice`, `checkpoint-splitover-integration` (standalone and combined with `checkpoint-main-slice`, both orders).
- Root cause looks like shard-level test isolation: under shard load, global `spawnRef`/`isSystemSpawned` state gets perturbed and `tryStartCheckpointWriter` takes the `"skipped"` branch instead of `"started"`.

PR #1471, opened 2026-06-30
fix(tui): patch @opentui/core@0.1.101 to fix CJK extmark drift without upgrading

yanyihan-xiaomi:
## Summary

Fixes the CJK paste-highlight drift bug (same issue addressed by #1385) by applying a targeted `bun patch` to `@opentui/core@0.1.101` instead of upgrading to 0.3.4.

This approach avoids the markdown rendering regression introduced by the 0.3.4 upgrade (streaming flicker worsening documented in #1385's risk section), while still fixing the core CJK extmark issue.

修复中文粘贴高亮偏移问题（与 #1385 相同的 bug），但采用 `bun patch` 修补 0.1.101 而非升级到 0.3.4，避免升级带来的 markdown 流式渲染回退。

## Root Cause

opentui 0.1.101's `ExtmarksController` uses incorrect values for extmark position adjustments:
- `insertText`: passes `text.length` (UTF-16 length) — CJK chars are length 1 but display-width 2
- `insertChar`: passes hardcoded `1` — wrong for any wide character
- `deleteCharBackward` / `deleteChar`: passes hardcoded `1` for deletion length

This causes paste highlights (`[Pasted ~N lines]`) to drift left by 1 column per CJK character typed before them.

Verified via headless reproduction: after inserting "中" at offset 0, `adjustExtmarksAfterInsertion` received `length=1` (should be 2); extmark shifted by 1 instead of 2.

0.1.101 的 extmarks 控制器在插入/删除时用字符串长度（UTF-16）和硬编码 1 代替显示宽度来偏移 extmark。每输入一个中文字符，粘贴高亮就左偏一列。通过无头测试复现确认。

## Fix

JS-level `bun patch` on `@opentui/core@0.1.101` — no native binary change needed:
- `insertChar`: `Bun.stringWidth(char)` instead of `1`
- `insertText`: compute display width per code point (newline=1, else `Bun.stringWidth`)
- `deleteCharBackward`: use `currentOffset - getVisualCursor().offset` after deletion (cursor position is correct post-deletion)
- `deleteChar`: compute width of char at cursor position from buffer text before deletion; also fixes the early-return guard which compared display-width offset against UTF-16 `getText().length` (caused the fix to be skipped entirely when cursor was past the string length but within display width)

Also pins `@opentui/core` overrides to `0.1.101` in root `package.json`.

通过 JS 层 patch 修复，用 `Bun.stringWidth` 计算实际显示宽度替代原有的 `text.length` 和硬编码 1。删除路径用光标差值（删除后光标位置正确）。同时修复了 deleteChar 的提前返回守卫（原代码用显示宽度与 UTF-16 长度比较，导致光标在 CJK 文本靠后位置时修复被跳过）。

## Why not upgrade to 0.3.4 (as in #1385)?

PR #1385 documented that the 0.3.4 upgrade itself **worsened streaming markdown flicker** — the `internalBlockMode="top-level"` prop mitigates but does not fully explain or resolve the underlying regression. Since opentui is the entire TUI rendering substrate, a major version jump carries risk of undiscovered side effects in scrolling, dialogs, IME, and syntax highlighting.

This patch isolates the fix to exactly the 4 broken code paths, with zero risk of markdown or rendering regressions. When opentui's markdown streaming behavior stabilizes in a future version, the patch can be removed and the version upgraded.

PR #1385 记录了升级 0.3.4 本身会加剧 markdown 流式闪烁，且底层渲染行为变化可能影响滚动、弹窗、输入法等未逐一验证的路径。本 patch 精确修复 4 处有 bug 的代码，零 markdown 回退风险。未来 opentui markdown 渲染稳定后可去除 patch 并升级。

## Verification

- Headless test confirms all 4 paths now use correct display width (insertChar "中" → length=2 ✓)
- 7 regression tests in `test/cli/cmd/tui/extmark-cjk.test.ts` (including the deleteChar guard edge case)
- All 34 TUI tests pass
- Full monorepo typecheck (18 packages) passes

## Known Limitations (not fixed, non-regression)

Multi-codepoint emoji (ZWJ sequences like 👨‍👩‍👧, skin-tone modifiers like 👍🏽) have incorrect width calculation in the per-codepoint sum approach. This is pre-existing behavior (the old `text.length` code was equally wrong for these cases) and not a regression from this patch. A proper fix would require grapheme-cluster-aware width calculation, which conflicts with opentui's internal per-codepoint cursor model. Similarly, `deleteLine` still uses UTF-16 length (also pre-existing, not in this patch's scope).

多码点 emoji（ZWJ 序列如 👨‍👩‍👧、肤色修饰如 👍🏽）的宽度计算仍不正确。这是 opentui 既有行为，非本次引入的回退——旧代码对这些 emoji 同样计算错误。正确修复需要 grapheme-cluster 级宽度计算，与 opentui 内部逐码点光标模型冲突，留作 follow-up。

## Changes

- `patches/@opentui%2Fcore@0.1.101.patch` — CJK extmark fix (insert/delete width + deleteChar guard)
- `package.json` — register patch + align overrides to 0.1.101
- `bun.lock` — updated for patched dependency
- `packages/opencode/test/cli/cmd/tui/extmark-cjk.test.ts` — 7 regression tests


qiaozongming:
### Review note — empirical check on the emoji-width limitation

The `## Known Limitations` section is honest about multi-codepoint emoji being unfixed, but the **technical attribution is incorrect** (it's inherited from #1385's claim), and I think it's worth correcting before the wrong model gets locked in by a "prevent regression" test.

I measured the actual editor visual-cursor delta against the various width computations by inserting each grapheme into a real `EditBuffer`/`EditorView` + `ExtmarksController`:

| input | **editor cursor delta (ground truth)** | this patch's per-codepoint sum (`\|\|1`) | per-codepoint sum w/o fallback | whole-string `Bun.stringWidth` |
|---|---|---|---|---|
| `👨‍👩‍👧` | **2** | 8 | 6 | 2 |
| `👍🏽` | **2** | 4 | 4 | 2 |
| `中` | 2 | 2 | 2 | 2 |

Two takeaways:

1. **The editor places the caret using true grapheme display width (= 2 for these emoji), not 6.** #1385's description states the editor treats `👨‍👩‍👧` as width 6 and adds a regression test to "prevent reverting to grapheme" — but the measured cursor delta is **2**, matching whole-string `Bun.stringWidth`, not 6 and not the per-codepoint sum of 8. The per-codepoint approach is the one that diverges from the editor here.

2. **Whole-string `Bun.stringWidth(text)` is actually correct for these emoji**, so why not use it? Because it counts `\n` as width 0 while the editor counts a newline as 1 column:

   | input | whole `Bun.stringWidth` | editor needs |
   |---|---|---|
   | `你\n好` | 4 | 5 |
   | `a\nb` | 2 | 3 |

   Multi-line paste always contains `\n`, so the per-codepoint loop exists specifically to special-case `\n`→1. That motivation is sound but isn't spelled out in the description.

**Suggested proper fix (follow-up, non-blocking):** iterate to special-case `\n`→1 / `\t`→2, but for non-newline runs use grapheme/whole-string `Bun.stringWidth` instead of summing per codepoint with `|| 1`. That fixes ZWJ sequences, skin-tone modifiers, combining marks, and variation selectors in one shot while preserving correct newline handling.

Everything else in this PR checks out: the deleteChar guard fix is correct (verified `中中中` + cursor at display-offset 4 now adjusts the extmark to `4,14` instead of being skipped), the lockfile catalog is now `0.1.101` with no drift on a plain `bun install`, all 7 regression tests pass, and `--frozen-lockfile` is clean. The "patch instead of upgrade" rationale is well-supported — #1385 (still a draft) documents real, unexplained streaming-markdown flicker regressions from the 0.3.4 upgrade. No objection to merging for its stated goal; the emoji case is a fair follow-up.

PR #1486, opened 2026-07-01
fix(actor): scope forkContext by session to prevent cross-session collisions

onlyfeng:
## Summary
`Actor`'s `forkContexts` map is keyed by `actorID` alone, but `actorID` is not globally unique across sessions — this lets one session's frozen fork context overwrite or be deleted by another's. This PR scopes the map key to `(sessionID, actorID)`.

## Root Cause
`packages/opencode/src/actor/spawn.ts:201` declares:
```ts
const forkContexts = new Map<string, ForkContext>()
```
keyed only by `actorID`, and `getForkContext` (interface at `spawn.ts:179`, implementation at `spawn.ts:712-713`) looks it up the same way.

`actorID` for subagents comes from `allocateActorID(sessionID, agentType)` in `packages/opencode/src/actor/registry.ts:314-336`. That function queries existing actor rows filtered by `session_id = sessionID AND agent = agentType` (`registry.ts:318-326`) and returns `${agentType}-${max + 1}` — i.e. the incrementing suffix is scoped per `(sessionID, agentType)`, not global. The first `explore` subagent spawned in *any* session gets `actorID = "explore-1"`.

Because `forkContexts` is a single process-wide `Map` keyed by `actorID` alone, two unrelated, concurrent sessions that each spawn their first `explore` subagent both get `actorID = "explore-1"` and collide on the same map entry:
- `spawn.ts:661` (`forkContexts.set(actorID, input.forkContext)` in the subagent path) — whichever session spawns second overwrites the first session's frozen `ForkContext` (system prompt + inherited messages + tools).
- `spawn.ts:577` / `spawn.ts:588` (cleanup on success/failure) and `spawn.ts:709` (`cancel`) all call `forkContexts.delete(actorID)` — cancelling or completing one session's actor deletes the *other* session's still-in-use context.

The consumer at `packages/opencode/src/session/prompt.ts:2930` (`spawnRef.current?.getForkContext(...)`) then reads whichever context happens to still be in the map, so a fork agent in one session can silently pick up (or lose) another unrelated session's frozen context snapshot.

## Changes
- `packages/opencode/src/actor/spawn.ts`: added a `forkContextKey(sessionID, actorID)` helper and routed every `forkContexts.set/get/delete` call through it; changed `getForkContext`'s signature (interface + implementation) to `(sessionID, actorID)`, matching the existing `cancel(sessionID, actorID, mode)` pattern.
- `packages/opencode/src/session/prompt.ts`: updated the sole call site to pass `sessionID` (already in scope) alongside the actor id.
- `packages/opencode/test/actor/spawn.test.ts`: updated existing `getForkContext` calls to the new two-arg signature; added a regression test that spawns same-`agentType` subagents in two different sessions, asserts they allocate the same `actorID` (the collision precondition), and asserts each session reads back only its own `forkContext` (including after cancelling one of them).
- `packages/opencode/test/inbox/fork-agent-compat.test.ts`: updated `getForkContext` calls to the new two-arg signature.
- `packages/opencode/test/session/classify-integration.test.ts`: updated the `spawnRef.current` mock's `getForkContext` stub to accept `(sessionID, actorID)` and match on `actorID` (it previously matched on the sole positional argument, which would silently receive `sessionID` under the new signature).

## Validation
- `bun typecheck`
- `bun test test/actor/spawn.test.ts test/inbox/fork-agent-compat.test.ts test/session/classify-integration.test.ts`

### DIFF
--- a/packages/opencode/src/actor/spawn.ts
+++ b/packages/opencode/src/actor/spawn.ts
@@ -176,7 +176,7 @@ export interface SpawnResult {
 export interface Interface {
   readonly spawn: (input: SpawnInput) => Effect.Effect<SpawnResult>
   readonly cancel: (sessionID: SessionID, actorID: string, mode: "graceful" | "forced") => Effect.Effect<void>
-  readonly getForkContext: (actorID: string) => Effect.Effect<ForkContext | undefined>
+  readonly getForkContext: (sessionID: SessionID, actorID: string) => Effect.Effect<ForkContext | undefined>
 }
 
 export class Service extends Context.Service<Service, Interface>()("@opencode/Actor") {}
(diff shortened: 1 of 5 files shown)


onlyfeng:
Closing to recreate from upstream/main. The forkContext cross-session collision bug still exists; will open a fresh PR with a rebased patch.

PR #1487, opened 2026-07-01
fix(session): honor toolChoice on max-mode's final step

onlyfeng:
## Summary
On the last step of a run, max-mode sessions can keep proposing and executing tool calls instead of being forced to wrap up with a text-only reply. This fixes the step-cap enforcement so it also applies when max-mode is active.

## Root Cause
In `packages/opencode/src/session/prompt.ts`, the main step loop computes `isLastStep = step >= maxSteps` and builds `processArgs` with `toolChoice: isLastStep ? "none" : ...` (around prompt.ts:3182) specifically so the final step forces a text-only wrap-up.

However, when max-mode is active (`useMaxMode`), the step is dispatched to `MaxMode.runMaxStep(...)` instead of `handle.process(processArgs)` (prompt.ts:3224-3235). `MaxStepInput.toolChoice` in `packages/opencode/src/session/max-mode.ts` (around max-mode.ts:61-66) is explicitly documented as "Accepted (so the same processArgs object can be spread in) but unused" — `runMaxStep` always runs its N parallel propose-only candidates and executes the winner's proposed tool calls via `handle.replay`, regardless of `toolChoice`. So on the last step, with max-mode on, the model can still propose and execute tool calls, silently bypassing the step-cap wrap-up.

## Changes
- `packages/opencode/src/session/prompt.ts`: gate the `stepEffect` ternary so `useMaxMode && !isLastStep` selects `MaxMode.runMaxStep`; the last step always falls back to `handle.process`, which does honor `toolChoice: "none"`.
- `packages/opencode/test/session/prompt-effect.test.ts`: add a regression test that pins the `max` agent to a single step (`agent: { max: { steps: 1 } }`) with max-mode enabled (`experimental: { maxMode: { candidates: 2 } }`), then asserts only a single LLM call happens on that step — proving `handle.process` ran instead of the 2-candidates+1-judge fan-out `runMaxStep` would otherwise trigger.

## Validation
- `bun typecheck`
- `bun test test/session/prompt-effect.test.ts -t "max mode's last step"`

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -3221,7 +3221,7 @@ NOTE: At any point in time through this workflow you should feel free to ask the
               return "break" as const
             }
 
-            const stepEffect = useMaxMode
+            const stepEffect = useMaxMode && !isLastStep
               ? MaxMode.runMaxStep({
                   // runMaxStep reuses the identical per-step args as handle.process,
                   // plus the orchestration handles it needs.
--- a/packages/opencode/test/session/prompt-effect.test.ts
+++ b/packages/opencode/test/session/prompt-effect.test.ts
@@ -301,6 +301,25 @@ function providerCfg(url: string) {
   }
 }
 
+// Enables the "max" agent (only registered when experimental.maxMode is
+// configured) and caps it at a single step, so the very first step is also
+// the last one.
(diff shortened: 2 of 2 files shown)


onlyfeng:
Closing to recreate from upstream/main. The max-mode last-step toolChoice bug still exists; will open a fresh PR with a rebased patch.

PR #1499, opened 2026-07-01
fix(tui): improve Task component rendering for async actor operations

yanyihan-xiaomi:
## Summary

- Fixed spawned (background) actors showing misleading "completed · 0ms" state while still running — now correctly displays spinner with live tool activity until the actor finishes
- Added contextual labels for different actor operations: `Background Explore Task`, `Waiting for`, `Cancelling`/`Cancelled`
- `wait`/`cancel` operations now resolve actor description from the registry instead of rendering as bare spinners

## Problem

When the LLM uses `actor spawn` + `actor wait` pattern:
1. Spawn tool part immediately becomes "completed" (spawn returns actor_id right away)
2. Task component only checked `props.part.state.status`, rendering the "completed" format with 0ms duration
3. `wait` tool parts had no `description` in their input, rendering as empty spinners

## Solution

The `Task` component now:
1. Looks up the actor's actual runtime status from `sync.data.actor` registry
2. Shows running state (spinner) when tool part is "completed" but actor is still running/pending
3. Resolves description from actor registry for operations without inline description (wait/cancel/status)
4. Displays clear labels per action type: `Background <Agent> Task`, `Waiting for`, `Cancelling`→`Cancelled`

## Result

img_v3_02136_32a064c7-53bf-46ad-a16c-64411316f2ag

img_v3_02136_6578b75c-82a1-43fa-b0e1-b15459083efg

### DIFF
--- a/docs/compose/plans/2025-07-01-fix-task-component-spawn-rendering.md
+++ b/docs/compose/plans/2025-07-01-fix-task-component-spawn-rendering.md
@@ -1,0 +1,162 @@
+# Fix Task Component Rendering for Spawned Actors
+
+> [!NOTE]
+> This document may not reflect the current implementation.
+> See the final report for up-to-date state:
+> [Final Report](../reports/task-spawn-rendering.md)
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use compose:subagent (recommended) or compose:execute to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Fix the `Task` TUI component so that spawned (background) actors are displayed with running/spinner state until the actor actually completes, rather than showing a misleading "completed" state with 0ms duration.
+
+**Architecture:** The `Task` component currently determines its visual state solely from `props.part.state.status` (the tool part's completion status). For `spawn` operations, this becomes "completed" immediately (spawn itself returned successfully), even though the spawned actor is still running. The fix adds a secondary check against the actor's actual runtime status from `sync.data.actor[sessionID]`. Additionally, `wait` tool calls (which have no `description`) should display context about what they're waiting for.
+
+**Tech Stack:** SolidJS (TUI rendering), TypeScript
+
+## Global Constraints
+
+- Run typecheck from `packages/opencode` via `bun typecheck`
+- Tests cannot run from repo root; run from `packages/opencode`
+- Avoid `any` types; rely on type inference where possible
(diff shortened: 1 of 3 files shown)


qiaozongming:
## Review — LGTM on substance

Clean, well-scoped fix. The core bug (spawned actors showing "completed · 0ms" while still running) is correctly addressed. No blocking issues; a few minor nits below.

### What's correct
- **`isRunning()` fix** (`index.tsx:2793`) — when the tool part is `completed` but the actor registry status is `running`/`pending`, it now returns `true` → spinner shows. This is the crux of the bug and it's right. Safe default: if `actorStatus()` is `undefined` (registry not yet synced), it falls back to `false` → old "completed" behavior, so no regression.
- **`actorEntry()` keying** (`index.tsx:2746`) — uses `props.part.sessionID` to index `sync.data.actor`. This matches how `actor.registered` events in `sync.tsx` key by the parent `sessionID`, which is where the tool part lives. Correct.
- **`actorId` fallback chain** — for `wait`/`cancel`/`status` ops the tool sets `metadata.actor_id` (snake_case) while spawn/run set `metadata.actorId` (camelCase). The PR reads camelCase `props.metadata.actorId` and falls back to `inputActorId()` (from `operation.actor_id`), which those ops always carry. So description/status resolution works for all actions.

### Minor nits (non-blocking)
1. **`inputAction()`/`inputActorId()` lack the flat-input fallback** (`index.tsx:2733`, `2738`). The existing `input()` memo handles both shapes: `raw.operation ?? raw`. The two new memos only read `raw?.operation?.action` / `.actor_id`. If a spawn input ever arrives un-wrapped (flat), `inputAction()` is `undefined` → header falls through to `${agent} Task` instead of `Background ${agent} Task`. Cosmetic only (loses the "Background" label); spawn still resolves the actor via `metadata.actorId`. Worth a one-line fallback for consistency.
2. **"Waiting for — …" header persists after the wait resolves** (`index.tsx:2819`). Once the `wait` part completes, `isRunning()` goes false and the `└ N toolcalls · duration` line appears, but the header still reads "Waiting for". Reads slightly odd for a finished wait. Pure UX polish.
3. **Brief flash on spawn** — spawn's part becomes `completed` immediately; if the `actor.registered` sync event hasn't landed yet, `actorStatus()` is `undefined` for a frame → momentary "completed · 0ms" before the spinner kicks in. It's reactive so it self-heals. Acceptable.

None of these change correctness.

PR #1508, opened 2026-07-01
fix: support Windows backslash paths in memory path parsing

### LINKED ISSUE #1455: Memory tool returns empty results - path outside memory layout on Windows (non-ASCII username path)
Memory tool search returns empty results for all queries. The root cause is that all memory files are skipped during reconciliation with "path outside memory layout" warning.

## Environment

- OS: Windows 11
- MiMoCode: 0.1.4 (server-proxy)

## Steps to Reproduce

1. Start MiMoCode on a Windows machine whose username contains non-ASCII characters (e.g. CJK)
2. Use memory tool to search for any keyword (e.g. "test")
3. Always returns "No matches"
4. Check logs and find "path outside memory layout, skipping" warnings for every memory file

## Logs

```
WARN  service=memory.reconcile path=C:\Users\测试用户\.local\share\mimocode\memory\projects\global\MEMORY.md path outside memory layout, skipping
WARN  service=memory.reconcile path=C:\Users\测试用户\.local\share\mimocode\memory\sessions\<session_id>\checkpoint.md path outside memory layout, skipping
WARN  service=memory.reconcile path=C:\Users\测试用户\.local\share\mimocode\memory\sessions\<session_id>\notes.md path outside memory layout, skipping
```

Note: The actual Windows username contains non-ASCII (CJK) characters. This is likely the root cause — the reconcile step may not handle Unicode paths correctly on Windows.

## Database Check

| Table | Records | Status |
|-------|---------|--------|
| memory_fts | 0 | Empty (bug) |
| memory_fts_idx_data | 2 | Very few |
| part | 9548 | Normal |
| message | 1673 | Normal |

The `memory_fts` table has 0 records — the FTS index was never populated.

## Expected Behavior

Memory tool should correctly identify memory file paths, build the full-text search index, and return search results.

## Actual Behavior

All memory files are skipped during reconciliation. `memory_fts` table is empty. Search returns no results for any query. Note: the files exist and are readable — `grep` tool can search them and `read` tool can access them without issues.

## Possible Cause

The path comparison logic in the reconcile step may fail when the Windows user profile path contains non-ASCII characters. The runtime path and the expected memory layout path might not match after normalization, causing all files to be classified as "outside memory layout."

## Workaround

Using `grep` and `read` tools as alternatives for memory search.

Ivaloz:
Fixes #1455 - Support Windows backslash paths in memory path parsing

### DIFF
--- a/packages/opencode/src/memory/paths.ts
+++ b/packages/opencode/src/memory/paths.ts
@@ -43,7 +43,8 @@ function detectType(key: string): MemoryType {
 }
 
 export function parsePath(absPath: string): MemoryLocator | null {
-  const m = absPath.match(/\/memory\/(global|projects|sessions)(?:\/([^/]+))?\/(.+)\.md$/)
+  // Support both forward slashes (Unix) and backslashes (Windows)
+  const m = absPath.match(/[/\\]memory[/\\](global|projects|sessions)(?:[/\\]([^/\\]+))?[/\\](.+)\.md$/)
   if (!m) return null
   const [, scope, idMaybe, keyRaw] = m
   const scope_id = scope === "global" ? "" : (idMaybe ?? "")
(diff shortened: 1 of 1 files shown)


Ivaloz:
Closing in favor of #1467 which has a cleaner implementation with normalizeSeparators() and type guards.

PR #1513, opened 2026-07-01
feat(tui): vision-aware image paste with non-vision fallback

wqymi:
## Summary

Makes TUI image paste vision-aware, guarantees **non-vision models never receive image bytes** (which cause hallucination), and gives the agent a discoverable, cost-aware way to pick a vision model when it needs to dispatch an image subagent.

### Vision-aware paste + fallback
- **Unify vision capability:** removed the redundant hardcoded `supportsImageInput` special-casing in `transform.ts`; `capabilities.input.image` is the single source of truth (models.dev already marks mimo-v2.5 / mimo-v2-omni image-capable). Kept one explicit `mimo-auto` override.
- **TUI paste routing:** vision model → existing base64 attach; non-vision model → spill clipboard image to a temp file and insert an `@file` reference (matching autocomplete's file-part shape) with a toast. Same fallback for pasting an image-file path. PDF/SVG unchanged. New `Clipboard.spillImage`. i18n across 7 locales.
- **Build-time notice:** `system.ts` injects a `<vision-capability>` block ONLY for non-vision models.
- **Read tool harness:** a non-vision model reading an image gets a warning (no base64); vision/PDF reads unchanged. `Effect.catchDefect` so an unresolvable model degrades to non-vision rather than crashing.

### Model discovery (so `--model` is usable)
- **`actor models` verb:** read-only subcommand listing available models, optionally `--vision`-filtered, count-capped.
- **`--model` description** points at `actor models` for valid values.
- **Non-vision hints name real vision models** + point to `actor models --vision`, in both the system prompt and the Read-tool warning.

### Cost-aware vision model selection
- **`vision_model` config knob** (analogous to `small_model`) lets the user pin which model handles image/vision subagent tasks.
- **`getVisionModel()` resolver:** explicit `vision_model` wins; otherwise a smart default over vision-capable models — **in-house (mimo/xiaomi) preferred, then cheapest by input cost, then name**. Fixes the prior behavior where an expensive model was recommended purely because its name sorted first alphabetically.
- **`actor models --vision` and the non-vision hints** use this same ordering / resolved model, so the recommended `--model` example is the sensible default, not an arbitrary one.
- Also cherry-picks a related fix: a non-empty `provider.models` config now acts as an implicit whitelist.

## Test Plan

- [x] `bun typecheck` clean (pre-push turbo, all tasks pass)
- [x] `bun test` on the touched suites → all pass (provider / actor-models / read-vision-harness / read / clipboard-spill)
- [x] Non-vision model + paste image → temp file + `@file` reference + toast; vision model → base64 attach unchanged
- [x] Non-vision model + `read` image → warning naming the resolved vision model + `actor models --vision`; vision model → attachment
- [x] Unresolvable/missing model → degrades to non-vision (no crash)
- [x] `actor models` lists all; `actor models --vision` filters + orders in-house-first/cheapest; `--limit` caps
- [x] `getVisionModel`: explicit `vision_model` wins; in-house beats cheaper foreign; cheapest when no in-house; undefined when none
- [x] Main model IS vision → no `<vision-capability>` injection

Plans under `docs/compose/plans/`: tui-paste-image-vision-fallback, actor-models-discovery, vision-model-selection.

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -1276,6 +1276,11 @@ export function Prompt(props: PromptProps) {
           }
         }
         if (mime.startsWith("image/") || mime === "application/pdf") {
+          if (mime.startsWith("image/") && !activeModelSupportsImage()) {
+            insertFileReference(filepath)
+            toast.show({ message: t("tui.paste.image.fallback_path"), variant: "info", duration: 3000 })
+            return
+          }
           const content = await Filesystem.readArrayBuffer(filepath)
             .then((buffer) => Buffer.from(buffer).toString("base64"))
             .catch(() => {})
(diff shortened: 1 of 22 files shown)


danhduykhan-code:
> ## Summary
> 
> Makes TUI image paste vision-aware. When the active model supports vision, images attach as before. When it does not, images degrade to a **file-path reference** instead of ever putting base64 bytes into a non-vision model's context (which causes hallucination). The guarantee is enforced end-to-end: at the paste boundary, at system-prompt build time, and at the Read tool.
> 
> - **Unify vision capability (A):** removed the redundant hardcoded `supportsImageInput` special-casing in `transform.ts`; `capabilities.input.image` is now the single source of truth (models.dev already marks mimo-v2.5 / mimo-v2-omni as image-capable). Kept one explicit `mimo-auto` override (free-tier routing alias absent from models.dev).
> - **TUI paste routing (B):** vision model → existing base64 attach; non-vision model → spill clipboard image to a temp file and insert an `@file` reference (matching autocomplete's file-part shape), with a toast. Same fallback when pasting an image-file path. PDF/SVG handling unchanged. New `Clipboard.spillImage` helper. i18n key added across all 7 locales.
> - **Build-time notice (C.1):** `system.ts` injects a `<vision-capability>` block for non-vision models only, telling them they can't see images — dispatch a vision subagent (`actor run ... --model <vision model>`) with the image path, or use `hexdump` for raw binary needs.
> - **Read tool harness (C.2):** a non-vision model reading an image gets a warning (no base64 attachment); vision models and PDF reads are unchanged. Uses `Effect.catchDefect` so an unresolvable model degrades to non-vision rather than crashing.
> 
> ## Test Plan
> 
> - [x] `bun typecheck` clean (all 12 turbo tasks pass on pre-push)
> - [x] `bun test test/cli/tui/clipboard-spill.test.ts test/tool/read-vision-harness.test.ts test/tool/read.test.ts` → 41 pass, 0 fail
> - [x] Non-vision model + paste clipboard image → temp file written, `@<file>` inserted, toast shown
> - [x] Vision model + paste clipboard image → base64 attachment unchanged
> - [x] Non-vision model + `read` on an image → warning, no attachment; vision model → attachment present
> - [x] Unresolvable / missing model → degrades to non-vision (no crash), covered by tests
> - [x] Non-vision model system prompt contains `<vision-capability>`; vision model does not
> 
> Plan: `docs/compose/plans/2026-07-01-tui-paste-image-vision-fallback.md`

PR #1528, opened 2026-07-02
 Fix: retry on HTTP 400 and exclude system sessions from `mimo -c`

JinyuXiang-Mimo:
### Changes

1. **Add HTTP 400 to retryable status codes** — `RETRYABLE_HTTP_STATUS` now includes 400, so transient 400 errors trigger exponential backoff retry instead of failing immediately.

2. **Exclude system sessions from `-c` (continue)** — `mimo -c` and `mimo run -c` previously picked up Auto Dream / Auto Distill sessions as the "most recent session" to continue. Now both the TUI and headless CLI paths skip system-spawned sessions.

3. **Centralize `isSystemSession()` helper** — Extracted from the inline check in `dialog-session-list.tsx` into `auto-dream.ts`. All three consumers (`run.ts`, `app.tsx`, `dialog-session-list.tsx`) now reference the same function. Adding a new system session type in the future only requires updating the `SYSTEM_SESSION_TITLES` set in one place.

4. **Suppress pre-existing SDK type error** — The `role` field is accepted server-side but missing from SDK codegen output. Added a type widening until the SDK is regenerated.

### DIFF
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -4,6 +4,7 @@ import { pathToFileURL } from "url"
 import { UI } from "../ui"
 import { cmd } from "./cmd"
 import { Flag } from "../../flag/flag"
+import { isSystemSession } from "../../session/auto-dream"
 import { bootstrap } from "../bootstrap"
 import { EOL } from "os"
 import { Filesystem, Log } from "../../util"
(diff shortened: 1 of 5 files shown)


danhduykhan-code:
> ### Changes
> 
> 1. **Add HTTP 400 to retryable status codes** — `RETRYABLE_HTTP_STATUS` now includes 400, so transient 400 errors trigger exponential backoff retry instead of failing immediately.
> 
> 2. **Exclude system sessions from `-c` (continue)** — `mimo -c` and `mimo run -c` previously picked up Auto Dream / Auto Distill sessions as the "most recent session" to continue. Now both the TUI and headless CLI paths skip system-spawned sessions.
> 
> 3. **Centralize `isSystemSession()` helper** — Extracted from the inline check in `dialog-session-list.tsx` into `auto-dream.ts`. All three consumers (`run.ts`, `app.tsx`, `dialog-session-list.tsx`) now reference the same function. Adding a new system session type in the future only requires updating the `SYSTEM_SESSION_TITLES` set in one place.
> 
> 4. **Suppress pre-existing SDK type error** — The `role` field is accepted server-side but missing from SDK codegen output. Added a type widening until the SDK is regenerated.
danhduykhan@gmail.com

PR #1536, opened 2026-07-03
fix: use target agent's configured model in plan_enter/plan_exit

### LINKED ISSUE #1526: fix(plan): plan_enter/plan_exit use getLastModel() instead of target agent's configured model
## Problem

`plan.ts` L71 (`plan_enter`) and L151 (`plan_exit`) use `getLastModel()` to determine the model for the synthetic user message created when switching modes. `getLastModel()` returns the model from the **previous** message in the session — which is the source agent's model, not the target agent's configured model.

### Example

1. User is in plan mode with `MiniMax-M3` (plan's configured model)
2. User calls `plan_exit` to switch to build mode
3. `plan_exit` creates a synthetic user message with `agent: "build"` and `model: getLastModel()` → returns `MiniMax-M3`
4. Build agent now uses `MiniMax-M3` instead of its configured `xiaomi/mimo-v2.5-pro`

**Result:** After `plan_exit`, the build agent uses plan's model instead of build's model.

## Root Cause

```typescript
// plan.ts L151 (plan_exit)
const model = getLastModel(ctx.sessionID) ?? (yield* provider.defaultModel())
```

`getLastModel()` iterates messages and returns the first user message's model — which is the plan agent's M3, not build's v2.5-pro.

## Proposed Fix

Use the target agent's configured model from `Agent.Service`:

```typescript
// plan_exit — target is "build"
const agentInfo = yield* Agent.Service.use((svc) => svc.get("build")).pipe(Effect.orDie)
const model = agentInfo.model
  ? { providerID: agentInfo.model.providerID, modelID: agentInfo.model.modelID }
  : getLastModel(ctx.sessionID) ?? (yield* provider.defaultModel())
```

Same fix for `plan_enter` (target is "plan").

## Files

- `packages/opencode/src/tool/plan.ts` — L71 (`plan_enter`) and L151 (`plan_exit`)

## Environment

- MiMo Code version: 0.0.0-main-202607021140
- Platform: macOS arm64

Ivaloz:
Fixes #1526 - Use target agent's configured model instead of getLastModel() which returns the previous message's model

### DIFF
--- a/packages/opencode/src/tool/plan.ts
+++ b/packages/opencode/src/tool/plan.ts
@@ -8,6 +8,7 @@ import { MessageV2 } from "../session/message-v2"
 import { Provider } from "../provider"
 import { Instance } from "../project/instance"
 import { type SessionID, MessageID, PartID } from "../session/schema"
+import { Agent } from "../agent/agent"
 import ENTER_DESCRIPTION from "./plan-enter.txt"
 import EXIT_DESCRIPTION from "./plan-exit.txt"
 
@@ -18,12 +19,20 @@ function getLastModel(sessionID: SessionID) {
   return undefined
 }
 
+function getAgentModel(agentInfo: Agent.Info) {
+  if (agentInfo.model) {
+    return { providerID: agentInfo.model.providerID, modelID: agentInfo.model.modelID }
+  }
+  return undefined
+}
+
 export const PlanEnterTool = Tool.define(
   "plan_enter",
   Effect.gen(function* () {
     const session = yield* Session.Service
     const question = yield* Question.Service
     const provider = yield* Provider.Service
+    const agentService = yield* Agent.Service
 
     return {
       description: ENTER_DESCRIPTION,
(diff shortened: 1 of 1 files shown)


yanyihan-xiaomi:
Not planned for now. See https://github.com/XiaomiMiMo/MiMo-Code/issues/1526#issuecomment-4913605026 for the rationale.

`plan_exit` uses `getLastModel()` intentionally to preserve the session's current model during agent-initiated switches. The user didn't make a new model selection — they only confirmed the switch. If they want a different model, they can dismiss the prompt and manually switch agent and model in the TUI.

In the typical flow (build → plan_enter → plan_exit → build), the build model is carried through via `getLastModel()` the entire time.

yanyihan-xiaomi (changes requested):
Not planned for now. Even if we were to adopt this approach in the future, the current implementation has a gap:

`getAgentModel()` only checks `agentInfo.model` but ignores `agentInfo.modelRef`. Agents can be configured via either:
- `model: { providerID, modelID }` — direct specification
- `modelRef: "ultra"` — a string reference resolved via `provider.resolveModelRef()`

The canonical path in `createUserMessage()` (prompt.ts:1507-1511) handles both:
```typescript
const agentModel = ag.modelRef
  ? yield* provider.resolveModelRef(ag.modelRef).pipe(...)
  : ag.model
```

Since `getAgentModel()` is a plain function (not an Effect), it cannot perform the async resolution needed for `modelRef`. If a user configures their agent with `modelRef` (which is likely the more common pattern), this function returns `undefined` and falls through to `getLastModel()`, making the entire change ineffective.

PR #1537, opened 2026-07-03
fix: mark all subagents as non-interactive for permission asks

Ivaloz:
All subagents (whether 'run' sync or 'spawn' background) are non-interactive: they have no attached human to reply to permission prompts, so an ask that would block must fail clean with DeniedError instead of hanging.

Previously, only SYSTEM_SPAWNED_AGENT_TYPES and background actors were marked non-interactive. This caused sync subagents (run operation) to hang indefinitely when they triggered a permission ask, because they were treated as interactive actors waiting for user confirmation.

This fixes the issue where subagents would hang forever waiting for permission confirmation that never comes.


Ivaloz:
Closing - conflicts with upstream, will rebase

PR #1631, opened 2026-07-08
fix: merge consecutive system messages for openai-compatible providers

yanyihan-xiaomi:
## Summary

Fixes #1614. Closes #1049. Supersedes #61.
Related: #1621.

vLLM/Qwen chat templates only accept a single leading system message, but MiMoCode produces multiple (agent prompt + memory instructions). This causes `"System message must be at the beginning"` errors for local deployments using `@ai-sdk/openai-compatible`.

## Changes

- In `ProviderTransform.normalizeMessages`, collapse consecutive leading system messages into one (joined with `\n\n`) for `@ai-sdk/openai-compatible` providers only
- Other providers (`@ai-sdk/openai`, `@ai-sdk/anthropic`, etc.) are unaffected
- Added design note explaining why `providerOptions` can be safely dropped at this pipeline stage

## Tests

6 new test cases covering:
- 2 system messages merged → 1
- 3 system messages merged → 1
- Single system message untouched (no-op)
- `@ai-sdk/openai` provider unaffected (scoping)
- Empty messages array (no-op)
- No system messages present (no-op)

All 168 tests in `transform.test.ts` pass.

PR #1636, opened 2026-07-08
fix(session): scope memory instructions to checkpoint owners; collapse system prompt to one message

wqymi:
## Summary

The LLM request assembled the system prompt as two `{role:'system'}` messages, and injected the memory-instructions block into every actor except a small hardcoded blacklist. That blacklist let the memory block leak into normal subagents that have no business seeing it, and the two-message split existed only for a cache-stability reason that doesn't actually hold within a session. This PR fixes the gating and collapses the system prompt to a single message.

## Memory-instructions gating

The gate keyed on `isSystemSpawned` (a blacklist of checkpoint-writer/dream/distill), so anything *not* in that set — including `explore`/`general`/compose subagents — received the memory block. But those subagents use per-actor compaction and have no checkpoint duty; teaching them to maintain `MEMORY.md`/`checkpoint.md` is at best noise and at worst misleading.

The correct audience is exactly the set the checkpoint flow serves: **main + peer**. This is the same judgement `SessionPrune.fireCheckpoints` already makes, so the two were duplicating (and could drift from) each other.

Introduces `ActorRegistry.servesCheckpoint(sessionID, agentID)` as the single source of truth, wired into both the memory gate (`LLM.buildSystemArray`) and the checkpoint self-trigger gate (`fireCheckpoints`). It keeps the two orthogonal exclusions explicit — agent **type** (system-spawned) and actor **mode** (subagent) — so a future system agent spawned as `mode:"peer"` can't silently slip back in.

| Actor | mode | Gets memory block? |
|-------|------|--------------------|
| main | main | yes |
| peer (orchestrator child session) | peer | yes |
| subagent (explore/general/compose) | subagent | no — was **yes** (the bug) |
| checkpoint-writer / dream / distill | subagent | no |

Fork capture is unaffected: it builds with the parent's `main` agentID, so the memory block stays in the fork prefix and the parent/fork byte-parity invariant still holds.

## Single system message

`buildSystemArray` emitted a stable base block plus a separate memory block so a plugin-appended part couldn't fragment the cache prefix. But within a session the memory block's paths are fixed, so the whole thing is byte-stable and caches identically as one message — the split only ever bought cross-session prefix sharing, which never happens (KV cache is per-session). One message also makes the fork-prefix parity invariant trivial and spares subagents/providers a stray extra system turn.

The `experimental.chat.system.transform` plugin hook still fires against the multi-part array (so plugins that index or append parts keep working); the collapse happens after. Blocks join with `\n\n` so adjacent markdown sections don't fuse into one heading. The old 2-part rejoin guard is removed.

## Test plan

- New `servesCheckpoint` coverage in `registry.test.ts`: main (undefined + "main"), peer, subagent, checkpoint-writer, unregistered (fail-open).
- Existing `prune-skip-system.test.ts` already exercises all four gate outcomes and passes unchanged against the shared helper.
- `llm-system-prompt.test.ts` asserts exactly one system message and that the memory content still lands for main.
- `fork-prefix-invariant.test.ts` / `llm-request-prefix.test.ts` confirm parent/fork system parity survives the collapse.
- `bun typecheck` clean; provider `transform.test.ts` (cache breakpoints) green.

### DIFF
--- a/packages/opencode/src/actor/registry.ts
+++ b/packages/opencode/src/actor/registry.ts
@@ -74,6 +74,7 @@ export interface Interface {
   readonly renderForAgent: (sessionID: SessionID) => Effect.Effect<string>
   readonly agentTypeFor: (sessionID: SessionID, actorID: string) => Effect.Effect<string>
   readonly isSystemSpawned: (sessionID: SessionID, actorID: string) => Effect.Effect<boolean>
+  readonly servesCheckpoint: (sessionID: SessionID, actorID: string | undefined) => Effect.Effect<boolean>
   readonly allocateActorID: (sessionID: SessionID, agentType: string) => Effect.Effect<string>
 }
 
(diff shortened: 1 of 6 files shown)


yanyihan-xiaomi:
This closes #1631

PR #1650, opened 2026-07-09
feat(provider,tui): default models to image-capable + per-model vision toggle

JinyuXiang-Mimo:
Flip the terminal fallback for capabilities.input.image from false to true in both model-build paths (fromModelsDevModel and the config merge): when neither models.dev metadata nor config declares modalities, a model is now assumed vision-capable instead of silently stripping images. A model is text-only only when it explicitly declares modalities.input without "image".

Adds a per-model vision toggle to the TUI model picker (ctrl+o, since ctrl+i collides with Tab in terminals): it rewrites the model's modalities.input in mimocode.json and live-reloads. Since vision is now the default, the row indicator flags only the minority that are text-only rather than tagging every vision model.

Updates provider tests: assert the new image=true default and text-only opt-out; the "prefers in-house vision model" test now asserts the winning provider rather than a specific id (more mimo models qualify as vision now).


yanyihan-xiaomi:
Not a good design

PR #1654, opened 2026-07-10
feat(tui): enable slash skills by default and add multi-skill orchestration

MiMoHardFather:
## Summary

- **Flip slash skills to enabled by default**: Replace `MIMOCODE_ENABLE_SLASH_SKILLS` with `MIMOCODE_DISABLE_SLASH_SKILLS` so skill-source commands appear in `/` autocomplete out of the box.
- **Add `/` trigger mid-message**: Extend autocomplete-detect to recognize `/skill-name` anywhere in the input (not just position 0), with guards against URL false positives and empty queries.
- **Multi-skill auto-loading**: When a user mentions 2+ `/skill-name` tokens in a single message, auto-load up to 3 SKILL.md contents and inject an orchestration plan prompt to guide the model.
- **i18n**: Add `tui.tips.multi_skills` tip across all 7 locales (en, es, fr, ja, ru, zh, zht).
- **Design docs**: Add Agent Multi-Skill Workflow Orchestration design documents in en/fr/ja/ru/zh.

## Test plan

- Type `/` at position 0 → legacy single-command behavior preserved
- Type `/skill-name` mid-message → autocomplete popup appears with skill matches
- Reference 2+ skills in one message → SKILL.md content auto-loaded + orchestration reminder injected
- Set `MIMOCODE_DISABLE_SLASH_SKILLS=1` → skills hidden from `/` autocomplete


yanyihan-xiaomi:
## Post-merge review: mid-message slash triggers have bugs — fix PR incoming

The multi-skill orchestration concept is sound and the range-aware `onSelect` is correct, but two pre-existing functions were not adapted for mid-message triggers:

### Issues found

1. **Input clearing**: `hide()` (autocomplete.tsx:513-522) unconditionally deletes from `(0,0)` to cursor when slash autocomplete is dismissed. This was designed for position-0 single-command mode, but now fires on mid-message triggers and clears the entire input. Repro: type `/compose `, then `/c` → input cleared.

2. **Premature dismiss**: Line 540 regex `/^\S+\s+\S+\s*$/` matches any 2-word input, so mid-message slash (e.g. `/compose /c`) immediately triggers dismiss → triggers bug 1 above. Users never reach the (correct) range-aware `onSelect`.

3. **No i18n for skill descriptions**: `slashCommandDescription` only translates 6 hardcoded commands. Skills in the `/` popup show raw English while `/skills` dialog uses `skillDescription` with proper i18n.

### Fix

Both bugs are caused by missing `store.index === 0` guards — the deletion in `hide()` and the dismiss regex should only apply to position-0 triggers. Fix PR adds these guards + skill i18n. Multi-skill orchestration and mid-message autocomplete preserved as-is.

Issue #1371, opened 2026-06-26
M3 (MiniMax-M3) model stringifies nested 'operation' envelope for tools with discriminatedUnion schema


ethanchen669:
# Bug Report: M3 model stringifies nested `operation` envelope for tools with `discriminatedUnion` schema

## Summary

When **M3 (MiniMax-M3)** is the primary agent and emits a tool call for a tool whose
`parameters` schema is a `z.strictObject({ operation: z.discriminatedUnion("action", [...]) })`
(e.g. `task`, `actor`, any future tool with the same shape), the model **stringifies
the entire `operation` envelope as a JSON string** rather than nesting an object. The
harness then fails to zod-parse the call. Affected tool calls surface to the user as:

```
⚙ invalid [tool=task, error=Invalid input for tool task: JSON parsing failed:
{"operation":{"action":"create","summary":"..."}}]
<]minimax[": "[
... more truncated invocations ...
... Unexpected EOF]
```

The downstream effect is that any primary agent configured as M3 (default in
`compose` mode and increasingly in custom setups) cannot reliably use `task`,
`actor`, or any tool with the same schema shape. This blocks the entire
specs-driven compose workflow.

## The M3-specific mutation

The same class of bug is already documented in
`packages/opencode/src/tool/task.ts` line 99-105 as a known defect for
`mimo-v2.5-pro`:

```typescript
// .meta({ type: "object" }) is REQUIRED — without it, the emitted JSON
// schema's `operation` node has only `anyOf`, no `type`. Some models
// (notably mimo-v2.5-pro) then stringify the entire envelope, producing
// {"operation":"{\"action\":\"create\",...}"} which fails zod validation.
// See research-tool-call-schema/REPORT.md §2.5 "success-nested" warning.
```

The `meta({ type: "object" })` annotation was added to `task` and `actor` to
mitigate this for those two tools. **M3 exhibits the identical bug**, and the
mitigation does not fully cover it because:

1. **M3 stringifies even when `type: "object"` IS present.** Empirically, even
   with the annotation, M3 produces:
   - `{operation: "{action:\"create\",summary:\"x\"}"}` (stringified, with bare keys)
   - Or wrapped in `<|object_ref_start|>minimax<|object_ref_end|>` markers
   - Output truncated mid-stream when the payload exceeds M3's per-call budget

2. **The bug is probabilistic.** Sometimes the call works. Sometimes the
   zod-parse fails. There's no clean error message — the agent just gets
[...]

PR #1665, opened 2026-07-11
fix(provider): accept discriminated-union tool schemas on Moonshot/Kimi

### LINKED ISSUE #488: Kimi K2.7 Code Not supported
### Description

I'm using opencode go as provider, other model is working fine, only kimi model is not working both for K2.7 Code and K2.6, here is the error message

```
Error from provider (Moonshot AI): Invalid request: tools.function.parameters is not a valid moonshot flavored json schema, details: <At path 'properties.operation': when using anyOf, type should be defined in anyOf items instead of the parent schema>
status 400 · non-retryable
```

Is there any additional parameter that need to be added for this model? Or is it just not supported ? If not supported will it be supported in the future?

### Plugins

_No response_

### OpenCode version

0.1.0

### Steps to reproduce

_No response_

_No response_

### Operating System

Ubuntu 24.04

### Terminal

_No response_

mvalentsev:
Fixes #488

Kimi (K2.7 Code, K2.6, K2.5) and the moonshot-v1-* models 400 on every request
once our tools are attached:

> tools.function.parameters is not a valid moonshot flavored json schema … when
> using anyOf, type should be defined in anyOf items instead of the parent schema

`task`/`actor`/`cron`/`session` build `operation` from a discriminated union
with `.meta({ type: "object" })`, so it serializes to
`{ "type": "object", "anyOf": [...] }`. Moonshot won't take a `type` next to
`anyOf` — it wants it inside each variant, which they already have. I can't just
drop the `.meta` (that's what stops MiniMax-M3/mimo from stringifying the whole
envelope, #1371), so I strip the parent `type` in `ProviderTransform.schema`
only for Moonshot/Kimi. Nothing else changes.

Verified against the real API (via a proxy so I could see the request):
kimi-k2.7-code hit that exact 400 before and returns 200 after, with the tool
loop running end to end. Every model on the endpoint (all 11) was broken and
works now. Added tests, including a guard that other providers keep the parent
type.

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -1277,6 +1277,56 @@ function flattenDiscriminatedUnion(schema: JSONSchema.BaseSchema | JSONSchema7):
   } as JSONSchema7
 }
 
+// Models served by Moonshot AI (Kimi). Matched on both the provider id
+// (moonshotai, moonshotai-cn, kimi-for-coding, …) and the model id (kimi-*), so
+// Kimi models reached through a gateway/proxy still get the schema fixup below.
+function isMoonshot(model: Provider.Model): boolean {
+  const provider = model.providerID.toLowerCase()
+  const apiID = model.api.id.toLowerCase()
+  return (
+    provider.includes("moonshot") ||
+    provider.includes("kimi") ||
+    apiID.includes("moonshot") ||
+    apiID.includes("kimi")
+  )
+}
+
+// Moonshot's "flavored" JSON-schema validator rejects a tool-parameter node that
+// carries a `type` as a sibling of an `anyOf`: it wants the type inside each
+// `anyOf` item instead. Our discriminated-union tool parameters
+// (task/actor/cron/session `operation`, declared as
+// `z.discriminatedUnion(...).meta({ type: "object" })`) serialize to
+// `{ type: "object", anyOf: [...] }`, so Moonshot 400s with:
+//   "when using anyOf, type should be defined in anyOf items instead of the parent schema"
+// Push the parent `type` into any item that lacks its own, then drop the parent
+// `type` — the variants already declare `type: "object"`, so this removes only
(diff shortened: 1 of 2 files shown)


yanyihan-xiaomi:
Thanks for the fix, merged!

One question: is this `type`-next-to-`anyOf` rejection a **Moonshot endpoint** restriction (their API gateway validator), or something inherent to the **Kimi model** itself? i.e. if Kimi is served through a different endpoint (OpenRouter, custom gateway, etc.), would the same rejection happen? Curious whether `isMoonshot` should be scoped to the endpoint or the model — would appreciate any info you have on this.

mvalentsev:
Hey @yanyihan-xiaomi, thanks for merging!

Short version: it's the endpoint — Moonshot's backend validator — not the Kimi model. I checked it three ways:

- **Moonshot directly** (`api.moonshot.ai`): the broken `{ type, anyOf }` tool schema 400s on every model I tried there (Kimi *and* the non-Kimi `moonshot-v1-*`), and 200s once the parent `type` is dropped.
- **OpenCode Go** (the setup from the report): ran it locally, same `kimi-k2.6`, same prompt — before the patch it dies with `Error from provider: Provider returned error` (Go just masks Moonshot's 400), after it replies and runs tool calls normally.
- **OpenRouter**, forcing the upstream for the same `kimi-k2.7-code` + same schema: `provider: { order: ["moonshotai"] }` → the "moonshot flavored json schema" 400, while `["deepinfra"]` / `["siliconflow"]` → 200.

So the same model is fine everywhere except when the request actually lands on Moonshot's backend. That's why `isMoonshot` matches on the model family rather than the provider id: gateways route Kimi into Moonshot under their *own* id (`opencode-go`, not `moonshotai`) — exactly this report — so a provider/baseURL gate would miss it. And where it isn't needed, the change is a no-op: it only drops a parent `type` the variants already declare, so those keep returning 200.

PR #1693, opened 2026-07-13
fix(session): stream transcript imports and skip oversized files

### LINKED ISSUE #1671: Bug: mimo crashes on launch (0xC0000409 / RADAR_PRE_LEAK_64) - Claude Code import reads large transcripts whole-file and OOMs
## Summary

`mimo` crashes immediately on launch and on `mimo --help` with exit code `0xC0000409` (STATUS_STACK_BUFFER_OVERRUN). Windows Error Reporting logs it as `RADAR_PRE_LEAK_64` (heap memory runaway). `mimo --version` works because it exits before the offending code path.

Root cause: the Claude Code session import runs unconditionally on every startup and reads each `~/.claude/projects/**/*.jsonl` file into memory with `readFile` + `String.split("\n")`. A single large transcript (in my case 2.5 GB; several others 100–370 MB) OOMs the process. The `try/catch` around the import can't catch a native OOM abort, so the app is permanently bricked.

## Environment

- MiMoCode: 0.1.5 (latest, installed via `mimo.xiaomi.com/install.ps1`). Also reproduces on 0.1.4.
- OS: Windows 11 25H2 (build 26200.8655) — **not OS-specific**; see the decisive test below.
- `~/.claude/projects/`: 3.9 GB across 465 `.jsonl` files; largest single file 2.5 GB.

## Reproduce

1. Have a large Claude Code transcript under `~/.claude/projects/<cwd>/<session>.jsonl` (mine is 2.5 GB).
2. Run `mimo --help` or `mimo` → no output, returns to prompt immediately.
3. Exit code: `-1073740791` (`0xC0000409`).

`mimo --version` returns `0.1.5` (exit 0).

## Evidence

**Launch log** (`~/.local/share/mimocode/log/*.log`) — every attempt ends at the same line:

```
INFO  service=server-proxy version=0.1.5 args=["--help"] process_role=main mimocode
INFO  service=db path=...mimocode.db opening database
INFO  service=db count=35 mode=bundled applying migrations
<crash - no further log line>
```

The crash lands right after DB migrations, which is where `ClaudeImport.run()` executes.

**Windows Event Log** (Application → Windows Error Reporting, event ID 1001):

```
EventName: RADAR_PRE_LEAK_64
P1: mimo.exe
P2: 1.3.14.0
P3: 10.0.26200.2.0.0
```

`RADAR_PRE_LEAK_64` = Windows heap-leak pre-warning → the process allocates memory without bound before aborting.

**Decisive test** — disable the Claude Code import:

```
$ MIMOCODE_DISABLE_CLAUDE_IMPORT=1 mimo --help
# prints full help, exit code 0
```

With the import disabled, `mimo` starts normally on the same machine. This pinpoints `ClaudeImport.run()` as the crash site.

**Ruled out** (all verified): binary corruption (fresh download hash-matches installed), AVX2 (CPU has AVX2; baseline build also crashes), PATH/config/home (crashes with an empty data dir too, because the input is `~/.claude/projects/`, not the MiMoCode DB), and Windows 25H2 (the import-disabled test runs fine on the same 25H2 machine).

## Root cause

`packages/opencode/src/session/claude-import.ts`:

```ts
// parse() - ~line 53
function parse(text: string, sessionId: SessionID) {
  const entries = text
    .split("\n")          // materializes the ENTIRE file into an array of lines
    .map((l) => l.trim())
    .filter(Boolean)
    .flatMap((line) => { try { return [JSON.parse(line)] } catch { return [] } })
  ...
}

// caller - ~line 275
const parsed = parse(await readFile(file, "utf-8"), sessionId)
```

`readFile` loads the whole `.jsonl` into one JS string, then `.split("\n")` materializes a second array of all lines, then each line is `JSON.parse`d. For a 2.5 GB file this peaks at ~5–8 GB and OOMs. The `try/catch` around `ClaudeImport.run()` in `index.ts` can't catch a native OOM abort.

`packages/opencode/src/index.ts` (~line 162) runs this on every launch, in the critical startup path:

```ts
if (!process.env.MIMOCODE_DISABLE_CLAUDE_IMPORT && !process.env.MIMOCODE_CLAUDE_IMPORTED) {
  process.env.MIMOCODE_CLAUDE_IMPORTED = "1"   // per-process env, not persisted -> re-runs every launch
  try { await ClaudeImport.run() } catch (e) { Log.Default.warn("claude-import failed", { e }) }
}
```

So a single oversized transcript permanently bricks the app — `mimo` can't start at all, with no size guard and no graceful failure.

## Suggested fix

1. **Stream, don't slurp**: read each `.jsonl` line-by-line (Bun file streaming / a line reader) instead of `readFile` + `split`. This alone fixes the OOM.
2. **Don't run on every startup**: make the import lazy (only on explicit `mimo session import` or first-run), or run it in a separate child process that can't crash the main launch, and/or skip files above a size threshold.
3. Even with (1), a size guard / memory budget would turn a future oversized file into a graceful skip instead of a hard abort.

## Workaround

Set `MIMOCODE_DISABLE_CLAUDE_IMPORT=1` (user env var). `mimo` then starts normally (without Claude Code history import). Archiving the oversized transcripts out of `~/.claude/projects/` also unblocks the import.

### Plugins

_No response_

### MiMoCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

fancyboi999:
### Issue for this PR

Closes #1671

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The Claude/Codex session importers read each `.jsonl` with `readFile` + `split("\n")`. A multi-GB transcript exceeds the JS string length limit and the process aborts natively — the `try/catch` around `ClaudeImport.run()` never sees it, so with the import in the startup path (until #1075 lands) the app is bricked by one oversized file.

Two changes, matching suggestions 1 and 3 from the issue:

1. **Stream, don't slurp** — `parse()` now consumes an (async) iterable of lines; the import path feeds it a line generator over `createReadStream` (new `session/jsonl.ts`). The file never exists as one string. `parse()` still accepts a plain string, so existing unit tests only gained an `await`.
2. **Size budget** — files over `MIMOCODE_IMPORT_MAX_FILE_BYTES` (default 512 MB, `0` disables) are skipped up front with a warn log, an `oversized` stat, and a CLI hint. This is needed even with streaming because one transcript's parsed messages are still held in memory for the single DB transaction — measured ~3.4 GB peak RSS for a 600 MB file — so an unbounded file would still exhaust memory, just later.

`codex-import.ts` gets the same treatment (identical `readFile`+`split` pattern on `~/.codex/sessions`). Deliberately **not** touching the startup call in `index.ts` — that's #1075's scope and this PR does not conflict with it.

Repro + before/after on macOS (32 GB RAM), 2.5 GB synthetic transcript in `~/.claude/projects/`:

| | `mimo --help` | peak RSS |
|---|---|---|
| main | exit 133 (SIGTRAP), no output, dies after `applying migrations` | 3.0 GB |
| this PR | exit 0, full help, transcript skipped as oversized | 548 MB |

- `session import-claude` prints `scanned 1, imported 0, resynced 0, skipped 0, oversized 1` plus the env-var hint.
- Real small session imports correctly end-to-end: verified session/message/part/external_import rows in SQLite, including `tool_result` back-patching onto the earlier `tool_use` part.
- 600 MB file with `MIMOCODE_IMPORT_MAX_FILE_BYTES=0`: streams through and imports fine (exit 0, 601 parts).
- `bun test test/session/jsonl.test.ts test/session/claude-import.test.ts test/session/codex-import.test.ts`: 33 pass. New tests cover chunk-boundary line splitting, CRLF, multi-byte UTF-8 across chunks, and string-vs-stream parse equivalence.
- `bun typecheck` clean; oxlint: no new warnings (the touched files actually dropped from 30 to 28).

### Screenshots / recordings

n/a (CLI)

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/session.ts
+++ b/packages/opencode/src/cli/cmd/session.ts
@@ -62,8 +62,11 @@ export const SessionImportClaudeCommand = cmd({
     const stats = await ClaudeImport.run({ force: args.force })
     UI.println(
       `Claude import: scanned ${stats.scanned}, imported ${stats.imported}, resynced ${stats.resynced}, skipped ${stats.skipped}` +
+        (stats.oversized ? `, oversized ${stats.oversized}` : "") +
         (stats.errors.length ? `, errors ${stats.errors.length}` : ""),
     )
+    if (stats.oversized)
+      UI.println(`${stats.oversized} transcript(s) exceeded the import size limit; set MIMOCODE_IMPORT_MAX_FILE_BYTES to raise it (0 disables).`)
     for (const err of stats.errors) UI.error(err)
   },
 })
--- a/packages/opencode/src/session/claude-import.ts
+++ b/packages/opencode/src/session/claude-import.ts
@@ -1,6 +1,5 @@
 import path from "path"
 import { existsSync } from "fs"
-import { readFile } from "fs/promises"
 import { Slug } from "@mimo-ai/shared/util/slug"
 import { Glob } from "@mimo-ai/shared/util/glob"
 import { Global } from "../global"
@@ -14,6 +13,7 @@ import { SessionTable, MessageTable, PartTable } from "./session.sql"
 import { ExternalImportTable } from "./external-import.sql"
 import { SessionID, MessageID, PartID } from "./schema"
 import { MessageV2 } from "./message-v2"
+import { Jsonl } from "./jsonl"
 
 const log = Log.create({ service: "claude-import" })
 
@@ -50,19 +50,11 @@ function stringifyToolContent(content: unknown): string {
 
 type Built = { info: MessageV2.Info; parts: { part: MessageV2.Part; time: number }[] }
 
-function parse(text: string, sessionId: SessionID) {
-  const entries = text
-    .split("\n")
-    .map((l) => l.trim())
-    .filter(Boolean)
-    .flatMap((line) => {
-      try {
-        return [JSON.parse(line)]
-      } catch {
-        return []
-      }
-    })
-  if (entries.length === 0) return undefined
+export async function parse(input: string | AsyncIterable<string>, sessionId: SessionID) {
+  // String input is kept for tests and small payloads; the import path streams
+  // lines so a multi-GB transcript never exists as one JS string (issue #1671).
+  const lines = typeof input === "string" ? input.split("\n") : input
+  let sawEntry = false
 
   let cwd: string | undefined
   let version: string | undefined
@@ -81,7 +73,16 @@ function parse(text: string, sessionId: SessionID) {
     current = null
   }
 
-  for (const entry of entries) {
+  for await (const raw of lines) {
+    const line = raw.trim()
+    if (!line) continue
+    let entry: any
+    try {
+      entry = JSON.parse(line)
+    } catch {
+      continue
+    }
+    sawEntry = true
     if (entry.type !== "user" && entry.type !== "assistant") continue
     const t = entry.timestamp ? Date.parse(entry.timestamp) : Date.now()
     if (!cwd && entry.cwd) cwd = entry.cwd
@@ -213,6 +214,7 @@ function parse(text: string, sessionId: SessionID) {
   }
   flushAssistant()
 
+  if (!sawEntry) return undefined
   return {
     cwd: cwd ?? "",
     version,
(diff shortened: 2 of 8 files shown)


fancyboi999:
Closing this for now at the author's request to keep the PR queue current.

The implementation branch will remain available. If the maintainers would like to revisit the large-transcript import fix, please comment here or reopen the PR. I am happy to rebase it onto the latest main branch, address feedback, and rerun the import regression and runtime validation.

PR #1702, opened 2026-07-13
fix(session): never send an assistant prefill — sweep trailing incomplete assistant regardless of age

wqymi:
## Summary

Fixes a real Bedrock 400 (hit live in-session): `This model does not support assistant message prefill. The conversation must end with a user message. (Service: BedrockRuntime)`.

**Root cause:** a conversation ending in an INCOMPLETE assistant message (no `time.completed`) is a trailing prefill — the next request ends on an assistant turn, which the Bedrock Converse API rejects with a non-retryable 400. `sweepOrphanAssistants` kept such an orphan for up to `ORPHAN_AGE_MS` (1h), so a freshly-abandoned trailing assistant leaked into the sent list.

**Fix (at the source, not content-dropping):** sweep a **trailing** incomplete assistant immediately regardless of age → `toModelMessages` skips it → the request ends on a user/tool turn. A COMPLETED reply is never touched (guarded by `time.completed`); a NON-trailing incomplete assistant keeps the 1h age gate so a mid-conversation in-flight retry chain isn't falsely swept.

This is the "harness never PRODUCES a prefill" approach — no content-dropping of completed replies, no provider/model-id guessing.

## Test
- `prompt-sweep.test.ts`: trailing incomplete assistant is swept regardless of age; non-trailing incomplete assistant keeps the age gate; completed assistant untouched. 4 pass.
- `bun typecheck` exit 0; invalid-output-continuation + llm-request-prefix suites pass (no regression).

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1944,11 +1944,30 @@ NOTE: At any point in time through this workflow you should feel free to ask the
       // 50 min) so a still-active in-flight request is never falsely
       // swept while its retry chain is making progress.
       const ORPHAN_AGE_MS = 3_600_000
-      for (const m of msgs) {
+      // Index of the last assistant message in send order. A TRAILING incomplete
+      // assistant (the conversation currently ends on an assistant that never
+      // stamped `time.completed`) is definitionally abandoned: nothing will ever
+      // complete it, and sending it as-is makes the request end with an assistant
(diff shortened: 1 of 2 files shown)


wqymi:
Closing as duplicate. #1703 is the primary prefill fix (retry-first: detect prefill-400 → fix trailing assistant → reactive retry). This PR's send-side sweep is superseded by #1703's approach per the retry-first direction.

PR #1710, opened 2026-07-13
fix(inbox): actor-notification wording reflects task status, not just process completion

wqymi:
## Summary

`renderActorNotification`'s `completed` branch always rendered `Background actor "..." completed. Status: <reportedStatus ∥ unknown>`, conflating two independent facts:

1. **Semantic mix-up** — `event.status === "completed"` describes the background actor's *process* ending cleanly, which is not the same as the child's *task* succeeding. A child can exit cleanly while its task was `failed`/`partial`/`blocked`, yet the user always saw "completed" and assumed success.
2. **"Status: unknown" noise** — `reportedStatus` is scraped from the child's `**Status**:` header; children that don't follow the format fell back to `unknown`, so the misleading `Status: unknown` line showed up constantly.

## Change

The top-line verb is now driven by `reportedStatus` (the real task outcome), not the process lifecycle:

| reportedStatus | wording |
|---|---|
| `success` | `... completed (success).` — the only affirmative case |
| `partial` / `failed` / `blocked` | `... finished with status: <s>.` — never implies success |
| none reported | `... finished.` — neutral, no `Status: unknown` line |

The process-level `failed`/`cancelled` branches are unchanged.

`parseActorNotification` is updated as the exact inverse. Its status enum went from `completed/failed/cancelled/stalled` to `success/partial/failed/blocked/finished/cancelled/stalled`, so render↔parse round-trips cleanly. The TUI card in `routes/session/index.tsx` maps the new statuses to icons/colors (success ✓, partial ◐, blocked ⊘, finished ⊙, etc.).

## Files
- `packages/opencode/src/inbox/render.ts` — render wording + parse inverse + status enum
- `packages/opencode/src/cli/cmd/tui/routes/session/index.tsx` — card status→style map
- `packages/opencode/test/inbox/parse-actor-notification.test.ts` — round-trip coverage for every branch

## Verification
- `bun typecheck` → EXIT 0
- `bun test test/inbox/` → 26 pass, 0 fail

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -1498,8 +1498,11 @@ function UserMessage(props: {
           // structured rows rather than raw <actor-notification> XML.
           const style = createMemo(() => {
             const s = note().status
-            if (s === "completed") return { icon: "✓", fg: theme.success, label: "completed" }
+            if (s === "success") return { icon: "✓", fg: theme.success, label: "completed" }
+            if (s === "partial") return { icon: "◐", fg: theme.warning, label: "partial" }
             if (s === "failed") return { icon: "✗", fg: theme.error, label: "failed" }
+            if (s === "blocked") return { icon: "⊘", fg: theme.error, label: "blocked" }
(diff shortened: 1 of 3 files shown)


wqymi:
Duplicate of #1711 (which also renames user-facing 'actor'->'sub-session').

PR #1719, opened 2026-07-14
exp: sticky agent mode — mode locking + permission-based skill scoping

yanyihan-xiaomi:
## Overview

Experimental branch exploring "mode is immutable within a session" interaction model. **Will not be merged to main** — for trying out and discussion.

### 1. Sticky Mode

After the first message in a session, the mode is locked:
- **Build / Plan** form a group — can freely switch between them (Tab cycles within group)
- **Compose** and other modes are isolated — cannot switch once entered
- `/new` creates a fresh session and restores free selection

Contextual toasts:
- Build/Plan user tries to switch to compose → "In this session, you can only switch between build, plan"
- Compose user presses Tab → "Cannot switch mode mid-session after entering compose mode"

### 2. Permission Replaces Hidden

- Removed `hidden: true` from all compose skills
- Configured via agent permission: defaults deny `compose:*`, compose agent explicitly allows
- `Skill.available()` no longer filters by hidden — relies entirely on permission system
- Compose skills naturally appear in compose agent's system prompt skill listing

### 3. Plan Tools Scoped to Build/Plan

- Defaults deny `plan_enter`/`plan_exit`, build and plan explicitly allow both (symmetric)
- No registry.ts changes needed — `llm.ts:resolveTools()` already uses `Permission.disabled()` to strip denied tools
- Compose and future agents get a clean tool list automatically
- Differs from pre-#1207 (asymmetric per-agent, caused cache issues) — here both build/plan see both tools

### 4. Removed composeSkillsBlock() Injection

- No more hardcoded `<available_skills>` XML block injection
- compose.txt cleaned: removed `{{compose_skills}}` placeholder
- Subagent guidance updated to "distill instructions" model

## Design Tradeoffs

| Decision | Rationale |
|----------|-----------|
| `set()` vs `userSwitch()` separation | Only guard user actions, don't block system behavior (session restore, plan tools) |
| Reactive `!!lastUserMessage()` | No manual lock/unlock — `/new` and `/session` work correctly by nature |
| `move()` skips blocked agents | Tab cycles within group, doesn't stop at first blocked entry |
| Self-switch always allowed | `current === target` → no-op, no false toast |
| Permission defaults deny + specific allow | Future agents auto-inherit deny, no per-agent config needed |
| Plan tools: symmetric in build/plan | No tool list mutation on build↔plan switch, stable prefix cache |

## ⚠️ Breaking Changes

[...]

PR #1725, opened 2026-07-14
feat: sticky agent mode — mode locking + permission-based skill scoping

yanyihan-xiaomi:
## Summary

Lock agent mode after the first message in a session, ensuring a stable system prompt (skills + tools) throughout the conversation. Supporting change: replace `hidden: true` with a permission-based scoping system for compose skills and plan tools.

## Motivation

Mode-switching mid-session causes prefix cache invalidation (the system prompt changes when skills/tools change). This is both a performance cost and a source of confusion — the model's context shifts underneath the user. By locking mode after the first message:

1. **Stable prefix cache** — system prompt is fixed for the entire session
2. **Clear UX contract** — the user commits to a mode at session start
3. **Clean permission model** — future agents auto-inherit deny rules, no per-agent configuration needed

### 1. Sticky Mode (TUI)

- `FREE_SWITCH_GROUP = ["build", "plan"]` — can freely switch between these
- All other modes (Compose, etc.) are isolated once entered
- `set()` (system) vs `userSwitch()` (user) separation — only user actions are guarded
- Reactive derivation from `lastUserMessage()` — `/new` and `/session` work correctly without manual lock/unlock
- Contextual toast messages when blocked

### 2. Permission-Based Skill and Tool Scoping

- Compose skills: `defaults deny compose:*`, compose agent explicitly allows
- Plan tools: `defaults deny plan_enter/plan_exit`, build+plan allow (symmetric)
- `Skill.available()` relies on permissions, not `hidden` flag
- Remove `composeSkillsBlock()` injection — skills appear naturally in the system prompt

## Design Decisions

| Decision | Rationale |
|----------|-----------|
| `set()` vs `userSwitch()` | Only guard user actions; system paths (session restore, CLI) are never blocked |
| Reactive `!!lastUserMessage()` | No manual lock state; handles all session transitions naturally |
| `move()` skips blocked agents | Tab cycles within group, doesn't stop at first blocked entry |
| Symmetric plan tools in build/plan | No tool list mutation on build↔plan switch |
| Permission defaults deny + allow | Future agents auto-inherit deny, no per-agent config needed |

## Breaking Changes (need discussion)

1. **Custom modes**: User-defined agents not in `FREE_SWITCH_GROUP` are treated as isolated. Would need group configuration support.
2. **`hidden: true` deprecated**: Skills relying on hidden must migrate to permission config.
3. **Plan tools hidden from compose**: Compose agent no longer sees `plan_enter`/`plan_exit`.

## Tests

- `bun typecheck` — clean
- `bun test test/agent/agent.test.ts` — 48/48 pass
- `bun test test/session/prompt-skill-mention.test.ts` — 8/8 pass
- `bun test test/permission` — 141/141 pass

## Open Questions

- Should `FREE_SWITCH_GROUP` be configurable (e.g., user-defined agents can declare their group)?
- Should there be a `/unlock` escape hatch for power users?
- Is one-session-one-mode too strict for workflows that genuinely need mode changes mid-conversation?

Supersedes #1719 (experimental branch, closed).

### DIFF
--- a/docs/compose/reports/sticky-agent-mode.md
+++ b/docs/compose/reports/sticky-agent-mode.md
@@ -1,0 +1,119 @@
+---
+feature: sticky-agent-mode
+status: proposed
+specs: []
+plans: []
+branch: feat/sticky-agent-mode
+---
+
+# Sticky Agent Mode — Final Report
+
+## What Was Built
+
+An experimental mode-locking system where agent selection is permanent for the duration of a session. Once a session has content (any user message), the user cannot switch to a different mode group. Build and Plan form a single free-switch group; all other agents (Compose, etc.) are isolated — once you're in Compose, you stay in Compose until `/new`.
+
+Three supporting changes enable a clean per-mode experience:
+
+1. **Permission-based skill scoping** — compose skills use `deny`/`allow` permissions instead of `hidden: true`
+2. **Plan tools scoped to build/plan** — `plan_enter`/`plan_exit` denied by default, allowed only for build/plan (keeping compose's tool list clean)
+3. **Removed `composeSkillsBlock()` injection** — compose skills appear naturally in the system prompt via `available(agent)`
+
+## Architecture
+
+### Sticky Mode (TUI)
+
+**File:** `packages/opencode/src/cli/cmd/tui/context/local.tsx`
+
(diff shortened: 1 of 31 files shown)


yanyihan-xiaomi:
@wqymi @MiMoHardFather Check this

PR #1728, opened 2026-07-14
fix(provider): guard non-array message content in part mapping (j.map is not a function)

wqymi:
## Summary

- Adds `normalizeContentArray()` at the entry of `ProviderTransform.message()` to ensure `msg.content` is always an array before any downstream transform calls `.map()` on it
- Fixes the runtime crash `"j.map is not a function"` when content is a string, object, or undefined (allowed by the `ModelMessage` type but not expected by provider adapter `.map()` calls)
- Adds `TypeError` recognition in `message-v2.ts:fromError()` before the catch-all `UnknownError` so programming defects like this are diagnosable in logs
- Adds 8 regression tests covering string, undefined, null, object, and mixed content inputs

## Root Cause

The `ModelMessage` type from the AI SDK allows `content: string | Array<...>`. Some code paths in provider adapters and the AI SDK internals call `.map()` on content assuming it is always an array. When content arrives as a string (e.g. from a malformed upstream message or provider-side transform), `.map` is not a function and the `TypeError` gets collapsed into an opaque `UnknownError` by the `fromError()` catch-all at `message-v2.ts:1152`.

## Fix

The fix normalizes `msg.content` to always be an array at the pipeline entry point (`ProviderTransform.message()`), so all downstream transforms can safely iterate. A string `"Hello"` becomes `[{ type: "text", text: "Hello" }]`; undefined/null/object become `[{ type: "text", text: "" }]`.

## Files Changed

- `packages/opencode/src/provider/transform.ts` — added `normalizeContentArray()` helper and applied it at pipeline entry
- `packages/opencode/src/session/message-v2.ts` — added `TypeError` case before `UnknownError` catch-all
- `packages/opencode/test/provider/transform.test.ts` — added 8 regression tests, updated 4 existing tests for normalized content format

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -456,6 +456,25 @@ function applyCaching(msgs: ModelMessage[], model: Provider.Model): ModelMessage
   return msgs
 }
 
+// Normalize msg.content to always be an array of parts. The ModelMessage type
+// allows `content: string | Array<...>`, but some code paths (provider adapters,
+// AI SDK internals) call `.map()` on content assuming it's an array. A non-array
+// (string, object, undefined) at that point produces "j.map is not a function".
+// This guard runs once at the pipeline entry so all downstream transforms can
+// safely iterate over content.
+function normalizeContentArray(msgs: ModelMessage[]): ModelMessage[] {
+  return msgs.map((msg) => {
+    if (Array.isArray(msg.content)) return msg
+    if (typeof msg.content === "string") {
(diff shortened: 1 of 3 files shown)


wqymi:
Reopening: this was closed alongside a force-push (same second in the timeline), not intentionally. The branch fix/llm-part-map-guard (764fb1e) still carries the j.map non-array-content guard diff (149 lines), which main does not fully contain (main lacks normalizeContentArray). This guard stops the 'j.map is not a function' crash.

PR #1730, opened 2026-07-14
fix(session): single-empty-call retry + schema-aware empty-step (replaces #1726)

wqymi:
Reopens work from #1726 (auto-closed by a rebase force-push). Empty tool calls trigger an immediate single retry (mirrors autoRetryTextToolCall) instead of threshold-halt; isEmptyStep is schema-aware so no-arg tools like plan_exit called with {} are not false-flagged.


wqymi:
Superseded by #1782, which reverts the entire empty-step guard (it mis-fired: flagged legit no-arg tool calls like list_apps{} as 'empty') and replaces it with call-preamble-leak detection (retry+suppress). The empty-step guard approach in this PR is abandoned. Closing.

PR #1777, opened 2026-07-17
fix(tool/tool_script): discourage overuse and stop escaping return values

yanyihan-xiaomi:
### 1. Description over-recommended the tool

The old description opened with *"Use this for BATCH tool operations"* and buried the "when NOT to use" list mid-document. Combined with a small-scale example (glob → read a handful of files → filter), it read as endorsing `tool_script` for routine multi-step work — including cases where 2 or 3 direct tool calls in parallel would have been strictly better (each call gets its own TUI title/output/metadata panel and its own agent-facing result; wrapping in `tool_script` collapses that into one aggregated string).

Rewritten:

- Opens with **"USE THIS TOOL SPARINGLY"**.
- Front-loads the "Do NOT use" list (single call, < ~5 same-tool calls, need to see per-step output, decision depends on prior step, side-effects needing per-step review).
- Quantifies the threshold: `~10+` similar calls where intermediate outputs don't need to be seen.
- Replaces the example with a 40-files → top-20 aggregation that actually demonstrates the hiding-intermediate-results value.

### 2. Return values were double-escaped

The old output packed the return value through `JSON.stringify(v, null, 2)` and inlined it into a plain-text envelope:

```
status: completed

Result:
"line1\nline2"
```

For string returns, this meant the model (and the TUI) saw literal `\n` byte sequences instead of real newlines — every user-facing "newline" in the return value was rendered as the two characters `\` and `n`.

Wrap the tool result in `<tool_script status="...">` XML blocks with `<return_value>`, `<logs>`, `<trace>` children (`<error_message>` on failure paths):

```
<tool_script status="completed">
<return_value>
line1
line2
</return_value>
<trace count="1">
- read → success [12ms]
</trace>
</tool_script>
```

- **Strings pass through verbatim** — newlines and quotes stay readable.
- **Non-strings** still go through `JSON.stringify(v, null, 2)` since JSON is their natural representation; the shape stays inspectable and the JSON output is embedded in `<return_value>` without a second layer of escaping.
- Error / budget-exceeded / timeout / cancelled / code-error paths all follow the same shape (`<error_message>` instead of `<return_value>`), so a downstream parser sees one consistent structure.
- `<return_value>` and `<error_message>` use distinctive names so common words in the returned content (the literal token `result` or `error`) don't blur the tag boundary. `<logs>` and `<trace>` keep their names — those closing forms are rare enough in real content that ambiguity is negligible.

### Follow-up (this PR): tag rename + accurate byte-length reporting

Two tweaks on top of the initial XML output:

- **Tag rename**: `<result>` → `<return_value>`, `<error>` → `<error_message>` (see rationale above). Outer `<tool_script>` wrapper stays — it's unlikely to collide with returned content.
- **Byte-length accuracy**: `budget_exceeded` message now reports `Buffer.byteLength(returnedText, "utf8")` instead of JS string `.length`, matching the gate above. Non-ASCII returns no longer under-report their size.

## Compatibility

The TUI renderer (`session/index.tsx:2133/2277`) reads `part.state.output` as a plain string and renders it as fenced text — so it treats XML as text and displays it faithfully. No renderer changes are needed; a follow-up PR could split `<return_value>` / `<logs>` / `<trace>` into separate collapsible panels but that's out of scope here.

## Tests

Only one assertion needed updating (`"Logs:"` → `"<logs>"`); every other test assertion was a substring match that survives the new format (`echo:a`, `caught:`, `→ error`, `[\n  2,\n  4,\n  6\n]` indentation, etc.).

`bun test test/tool/tool-script.test.ts` → **23 pass · 0 fail**. `bun typecheck` → **EXIT 0**.

### DIFF
--- a/packages/opencode/src/tool/tool-script.ts
+++ b/packages/opencode/src/tool/tool-script.ts
@@ -169,7 +169,7 @@ export const ToolScriptTool = Tool.define(
             return {
               title: "code too large",
               metadata: { status: "code_error", toolCalls: 0 },
-              output: `<tool_script status="code_error">\n<error>\ncode exceeds ${MAX_CODE_BYTES} bytes\n</error>\n</tool_script>`,
+              output: `<tool_script status="code_error">\n<error_message>\ncode exceeds ${MAX_CODE_BYTES} bytes\n</error_message>\n</tool_script>`,
             }
           }
 
@@ -205,7 +205,7 @@ export const ToolScriptTool = Tool.define(
             return {
               title: "transpile error",
               metadata: { status: "code_error", toolCalls: 0 },
-              output: `<tool_script status="code_error">\n<error>\n${transpiled.error}\n</error>\n</tool_script>`,
+              output: `<tool_script status="code_error">\n<error_message>\n${transpiled.error}\n</error_message>\n</tool_script>`,
             }
           }
 
@@ -329,7 +329,7 @@ export const ToolScriptTool = Tool.define(
             return {
               title: status,
               metadata: { status, toolCalls: trace.length },
-              output: `<tool_script status="${status}">\n<error>\n${message}\n</error>\n${logBlock}${traceBlock}</tool_script>`,
+              output: `<tool_script status="${status}">\n<error_message>\n${message}\n</error_message>\n${logBlock}${traceBlock}</tool_script>`,
             }
           }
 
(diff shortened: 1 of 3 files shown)


JinyuXiang-Mimo (approved):
LGTM

JinyuXiang-Mimo (approved):
LGTM

PR #1801, opened 2026-07-19
fix(opencode): validate custom tool arguments

### LINKED ISSUE #1800: Custom tool arguments bypass runtime validation through tool_script
## Summary

Custom tools can declare a Zod schema for their arguments, but calls made through `tool_script` can reach the custom tool's `execute` function without runtime validation.

Observe that the custom tool's execute function receives the invalid value.

Expected behavior
The argument schema should be validated before execute is called. Invalid arguments should return a recoverable validation error, and execute should not run.
Actual behavior
The custom tool adapter in packages/opencode/src/tool/registry.ts registers parameters: z.object(def.args), but calls def.execute(args as any, ...) directly without parsing the arguments first. packages/opencode/src/tool/tool-script.ts forwards script-call arguments into this path.
By contrast, built-in tools validate arguments with parameters.parse(args) before execution.
Impact
Malformed arguments can reach plugin/custom-tool implementation code. Tool authors cannot safely rely on the declared schema, which may lead to runtime errors or unexpected behavior.
Suggested fix
Validate arguments with the registered schema before calling def.execute(...). Return a recoverable validation error when validation fails.
Please add a regression test that invokes a custom tool through tool_script with invalid arguments and verifies that execute is not called.

### Plugins

_No response_

### MiMoCode version

_No response_

### Steps to reproduce

1. Register a custom tool with an argument schema such as:

   ```ts
   args: {
     path: z.string(),
   }
2. Invoke the custom tool through tool_script with an invalid argument:
{ "path": 123 }

3. Observe that the custom tool's execute function receives the invalid value.


然后把正文里重复的 `## Steps to reproduce` 那一节删掉即可，其他内容保留。

_No response_

### Operating System

_No response_

### Terminal

_No response_

floze-the-genius:
### Issue for this PR

Closes #1800

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Custom tools now parse their declared Zod argument schema before `execute` runs.
Invalid input uses the existing recoverable validation-error path, while parsed,
coerced, and stripped values are passed to the custom tool.

The regression invokes a real file-based custom tool through `tool_script` with
an invalid `path` value and verifies that its `execute` side effect never runs.
This is scoped to runtime argument validation; it does not change the tool
sandbox or permissions model.

Rebased exactly across 105 upstream commits onto current `main` (`eef8206d`).
The patch-id is unchanged and the final diff still contains only
`registry.ts` plus its regression test.

- Bun 1.3.14 pre-push workspace typecheck: 12/12 tasks passed
- `bun test test/tool/registry.test.ts --timeout 60000`: 6/6 passed
- `bun test test/tool/tool-script.test.ts --timeout 60000`: 49/49 passed
- package typecheck: passed
- full Oxlint: zero errors; targeted changed files have zero errors and five
  pre-existing warnings in `registry.ts`
- `git diff --check`: passed

The repository GitHub workflows remain the required independent Linux/Bun
validation gate.

### Screenshots / recordings

Not applicable; this changes backend tool-dispatch validation and has no visual
surface.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -75,6 +75,7 @@ import { BuiltinWorkflow } from "@/workflow/builtin"
 import { ToolScriptTool, renderToolScriptDeclarations } from "./tool-script"
 import { toolScriptRegistry } from "./tool-script-ref"
 import { usesGPTToolset } from "./gpt"
+import { RecoverableError } from "./recoverable"
 
 const log = Log.create({ service: "tool.registry" })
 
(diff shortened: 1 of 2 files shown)


floze-the-genius:
Rebased onto current main (223 upstream commits) and adapted the regression to the current `exec` tool name. Validation: `bun test --timeout 30000 test/tool/registry.test.ts` (6 passed), package typecheck, and the Bun 1.3.14 pre-push workspace typecheck (12/12 tasks).

floze-the-genius:
@JinyuXiang-Mimo I exact-rebased this onto current main (941e088f, 0 behind; patch-id unchanged). The real file-based regression verifies that invalid tool_script arguments never reach custom-tool execution. On Bun 1.3.14, registry tests pass 6/6, tool-script tests 49/49, workspace typecheck 12/12, and Oxlint reports 0 errors. The test, typecheck, and lint workflows are waiting for maintainer approval. Could you approve the workflows and review when convenient?

PR #1876, opened 2026-07-23
feat: enable direct messaging to subagents from TUI

cetagostini:
## Summary
- Add `POST /:sessionID/inbox/send` server endpoint that wraps `Inbox.send()` for direct subagent messaging
- Add `onCustomSubmit` prop to `Prompt` component allowing custom submit routing (skips model/workspace checks)
- Show input bar when viewing a running, pending, or idle subagent — users can now type directly to subagents without relaying through the main agent

## Motivation
When navigating to a subagent's conversation view in the TUI, the input bar was hidden. All communication had to go through the main agent via `actor send`, adding latency. This PR enables direct messaging by:
1. Exposing the existing `Inbox.send()` infrastructure as an HTTP endpoint
2. Routing the TUI prompt's submit to that endpoint when viewing a subagent

If a sub-agent goes away from the main goal, you can provide small notch to put it back on track. If main agent miss indications, user can provide without let the agent go crazy.

### Server (`packages/opencode/src/server/routes/instance/session.ts`)
- New `POST /:sessionID/inbox/send` endpoint accepting `{ actorID, content, type? }` and returning `{ inboxID }`
- Reuses the same `inboxServiceRef.current.send()` that the `actor send` tool uses

### TUI Prompt (`packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx`)
- New `onCustomSubmit` prop on `PromptProps`: when provided, the submit function calls it instead of the default `session.prompt` path
- The custom path skips model selection, workspace checks, and slash command handling — just sends the text and clears the input

### TUI Session (`packages/opencode/src/cli/cmd/tui/routes/session/index.tsx`)
- New `subagentInputVisible` memo: returns `true` when viewing a subagent with status `running`, `pending`, or `unknown` (idle-without-outcome)
- Renders a second `<Prompt>` with `onCustomSubmit` that POSTs to the inbox send endpoint
- `SubagentFooter` (navigation + status) remains visible above the input bar

## Testing
1. Spawn a persistent peer via `session create --topic test`
2. Navigate to the child session in the TUI
3. Verify the input bar appears alongside the subagent footer
4. Type a message and submit — verify it arrives in the subagent's inbox and triggers a turn
5. Verify the subagent's response appears in its conversation view
6. Navigate back to main — verify the main agent's input bar works normally
7. Test with ephemeral subagents — verify input is disabled after terminal

## Out of Scope
- Backend Inbox changes (already complete)
- Permission system changes (handled by PR #1700)
- SDK regeneration (using direct fetch instead)

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -63,6 +63,9 @@ export type PromptProps = {
     normal?: string[]
     shell?: string[]
   }
+  /** When set, overrides the default session.prompt submit flow. The function
+   *  receives the trimmed input text and should return true on success. */
+  onCustomSubmit?: (text: string) => Promise<boolean>
 }
 
 export type PromptRef = {
@@ -1018,6 +1021,23 @@ export function Prompt(props: PromptProps) {
     if (props.disabled) return false
     if (autocomplete?.visible) return false
     if (!store.prompt.input) return false
+
+    // Custom submit path (e.g. subagent inbox send). Skip model/workspace checks.
+    if (props.onCustomSubmit) {
+      const trimmed = store.prompt.input.trim()
+      if (!trimmed) return false
+      const ok = await props.onCustomSubmit(trimmed)
+      if (ok) {
+        history.append({ ...store.prompt, mode: store.mode })
+        input.extmarks.clear()
+        setStore("prompt", { input: "", parts: [] })
+        setStore("extmarkToPartIndex", new Map())
+        props.onSubmit?.()
+        input.clear()
+      }
+      return ok
+    }
+
     const agent = local.agent.current()
     if (!agent) return false
     const trimmed = store.prompt.input.trim()
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -1026,16 +1026,20 @@ export function Prompt(props: PromptProps) {
     if (props.onCustomSubmit) {
       const trimmed = store.prompt.input.trim()
       if (!trimmed) return false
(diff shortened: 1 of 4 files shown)


cetagostini:
@wqymi and @MiMoHardFather  what do you think?

wqymi:
Nice work on this — overall LGTM! 🎉 The design makes sense and I like how you separated concerns: the input was hidden because the default submit only ever routes to the main agent, so a dedicated inbox endpoint + onCustomSubmit is exactly the right call. Server side mirrors the actor send tool cleanly, and the test coverage is solid.

One thing to fix in the test plan, plus a few polish items:

The test plan tests the wrong path. session create --topic spawns a peer (its own session, where it's main), and it's also an experimental Orchestrator-only tool (MIMOCODE_EXPERIMENTAL_ORCHESTRATOR), so following those steps the input bar never appears — subagentInputVisible() only fires for same-session subagents (currentAgentID() !== "main"). Could you retest with an actual subagent? i.e. have the main agent delegate a task to spawn a subagent, then use the "view subagents" keybind to navigate into it while it's running/idle — that's the path this PR actually ships, and where the input bar should show up.

Polish nits:
1. disabled={disabled()} on the subagent <Prompt> is dead code. subagentInputVisible() already returns false whenever there are pending permissions/questions — the exact condition disabled() checks — so it's always false here. Safe to drop.
2. The endpoint doesn't declare a 500. When the inbox service isn't running it fails with NamedError.Unknown → HTTP 500, but responses only lists 200/400/404. Mind adding 500 so the OpenAPI doc matches reality?
3. The failure toast shows the raw response body. Send failed: ${res.text()} dumps {"name":"NotFoundError",...} at the user — parsing out message/data.message reads nicer.
Also, heads up the branch is ~140 commits behind main — could you rebase and get CI green? It merges cleanly right now, so mostly just keeping it that way.

Solid PR — nothing blocking, just the test-plan fix + polish. Thanks for taking this on! 🙏

PR #1898, opened 2026-07-24
fix(tui): prevent keybinds from getting permanently suspended

### LINKED ISSUE #1897: fix(tui): Ctrl+X keybindings stop working until user types and clears text
## Description

Ctrl+X keybindings (the leader key system) sometimes stop working entirely. The user must type random text and clear it before shortcuts work again.

## Root cause

The command keybinds suspension mechanism uses a counter (suspendCount) that can get out of sync. Two independent suspension sources (ghost text prediction and autocomplete) both call keybinds(false) to suspend and keybinds(true) to resume. If their increments/decrements don't balance, keybinds stay permanently suspended.

Specifically:
1. Autocomplete show() calls keybinds(false) every time, even if already visible
2. Ghost text effect uses onCleanup which can fail to fire in race conditions

## Fix

1. Make autocomplete show/hide idempotent - only suspend/resume when visibility state actually changes
2. Replace onCleanup-based ghost text suspension with explicit state tracking

## Verification

- All 392 provider tests pass
- All 77 keybind/autocomplete tests pass
- TypeScript typecheck passes

anandlo:
Closes #1897

## What

Fix Ctrl+X keybindings stopping working until the user types random text and clears it.

## Why

The autocomplete component had no cleanup on unmount. When the Prompt component unmounts while autocomplete is visible (e.g. session switch, view toggle), `hide()` is never called, leaving the suspendCount stuck > 0. This prevents ALL command keybinds (including Ctrl+X leader combos) from firing.

Typing random text triggers onContentChange which can clear ghost text and/or dismiss autocomplete, resuming keybinds - hence the workaround.

## Changes

1. **autocomplete.tsx**: Make show/hide idempotent (skip suspend if already visible/hidden). Add `onCleanup` to resume keybinds on component unmount.
2. **prompt/index.tsx**: Reverted previous wrong fix. Ghost text effect already uses `onCleanup` correctly (fires on effect re-run AND on unmount). The bug was solely in autocomplete not cleaning up.

## Verification

- 77 keybind/autocomplete tests pass
- TypeScript typecheck passes
- Change is minimal: 2 files, 12 added, 11 removed

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/autocomplete.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/autocomplete.tsx
@@ -531,7 +531,7 @@ export function Autocomplete(props: {
   }
 
   function show(mode: "@" | "$" | "/") {
-    command.keybinds(false)
+    if (!store.visible) command.keybinds(false)
     setStore({
       visible: mode,
       index: props.input().cursorOffset,
@@ -539,7 +539,7 @@ export function Autocomplete(props: {
   }
 
   function hide() {
-    command.keybinds(true)
+    if (store.visible) command.keybinds(true)
     setStore("visible", false)
   }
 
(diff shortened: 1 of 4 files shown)


anandlo:
Closing this PR. The root cause analysis needs more investigation before implementation. Will revisit with a proper reproduction and regression test.

Issue #1837, opened 2026-07-20
[FEATURE]: Context occupancy meter (% + tokens) and adjustable auto-compact threshold (sidebar or settings)


ByJotaro:
﻿### Describe the enhancement you want to request

**Problem**

Context fill is hard to feel until overflow/rebuild. Auto-compact thresholds live in config (`compaction.*`, `reserved`, `limit.context`), but TUI has no:

- how full the window is (**% + tokens**);
- where the **auto-compact line** sits;
- a way to **nudge the threshold for this session** without editing JSON.

Related compaction correctness bugs (#1661, #725, #435) are separate — this is **visibility + session-level control**.

**Proposal (either UI is fine)**

**Option A — right sidebar / status column**  
- Bar: used/limit.  
- Label: `123k / 500k (24%)`.  
- Mark/line for compact/rebuild threshold.  
- Slider for **this session’s** effective threshold.

**Option B — dedicated settings screen**  
- Same metrics + slider/numeric field if sidebar is crowded.

Persist per session id (restored on resume); optional “save as default” to project/global config.

Slider should move the effective trigger consistently with `compaction.reserved` / checkpoint policy, with safe clamps so overflow is not guaranteed.

**Benefits**

- Users see when rebuild is coming.  
- Tune aggressiveness without config archaeology.  
- Aligns UI with existing compaction keys.

**Acceptance criteria**

- [ ] Live % and token counts.
- [ ] Compact threshold visible and matches real trigger within a small tolerance.
- [ ] Change is per-session and restored on resume of that session id.
- [ ] Does not rewrite global config unless “save as default”.
- [ ] Safe clamps.

**Environment**

- MiMoCode 0.1.7
- Windows 11 / Windows Terminal

PR #1930, opened 2026-07-27
feat(compaction): per-model early-compaction budget, and fix the Codex context clamp

yanyihan-xiaomi:
## Summary

Follow-up to #1926, which capped Codex `gpt-*` context at 300K. That fix did not work and had side effects:

- **It never moved the compaction trigger for the models it targeted.** `Overflow.usable()` reads `limit.input` whenever the catalog publishes one, and every 1M-class GPT model does (`gpt-5.6*`: `context 1_050_000, input 922_000`). Setting only `limit.context` changed the TUI percentage denominator and the MCP tool budget — compaction still fired at ~902K. The footer would read `900,000 (300%)` while the engine kept going.
- **It raised windows instead of lowering them.** Assignment, not clamp: `gpt-4o` 128K → 300K, `gpt-3.5-turbo` 16K → 300K, so those models were pushed past what the provider accepts.
- **It defeated a sentinel.** `gpt-image-*` publishes `limit.context: 0`, which `overflow.ts` uses to disable overflow handling entirely; 300K re-enabled it.

Root cause is that one field, `model.limit.context`, was carrying three different meanings: the provider's hard prompt capacity, the point where *we* choose to compact, and the denominator the UI displays. This PR separates them.

## What changed

**Provider capacity (`plugin/codex.ts`)** — now a clamp, applied only when a window exists, and `limit.input` is only clamped when the catalog already publishes one (introducing it would switch `usable()` to the input branch and silently drop the output reserve). The value is **372,000**, with sources in the comment: OpenAI's Codex model registry declares `context_window = max_context_window = 372000` for the gpt-5.6 variants (openai/codex#31860 quotes the served catalog), and a direct Codex request with 350,317 input tokens completes (can1357/oh-my-pi#5705). The 272,000 that Codex lowered its *bundled* metadata to (openai/codex#33972) is the `>272K input` 2x-input / 1.5x-output billing boundary — a spending policy, not capacity — so it ships as a documented config recipe instead.

**Working budget (new `compaction.max_context`)** — compact earlier than the model window, for cost, latency, cache-churn or answer-quality reasons. A token count, a `"300K"` / `"1M"` / `"50%"` shorthand, or a map keyed by `"<providerID>/<modelID>"` with wildcards. Always clamped to the provider cap, so it can only lower the trigger, never raise it. `0` means "no budget for this model". Addresses #1837.

**One source of truth** — `Overflow.contextWindow()` returns `{hard, effective, usable, source}`; `usable()` is a thin wrapper. With no budget configured the arithmetic reduces to the previous expression for both model shapes, verified by unit tests against the base file. Checkpoint thresholds and pruning follow the budget for free because they already call `usable()`.

**Display** — the prompt footer, subagent footer and sidebar previously divided by the raw `limit.context`, which never matched the trigger (off by the ~40K reserves). All three now divide by the trigger and mark a configured budget with `↓`: `33.0K/260K↓ (13%)`. `/status` gained a Context block (`window 1.05M · budget 300K (config) · reserved 40K · compacts at 260K`), and `mimocode models <provider>` prints the same numbers without `--verbose`.

**`/context-limit`** — preset picker (Model default / 200K / 300K / 500K / 1M / Custom…) writing the current model's key to the global config. It refuses while a session is busy, because a config write triggers `Instance.dispose()`, whose `SessionRunState` finalizer cancels every in-flight runner. It writes a single leaf key rather than read-modify-writing the map, so sibling entries survive and project-level budgets are not promoted into the user's global config.

## Impact by channel

`limit.input`/`limit.context` are per provider+auth, so the same model behaves differently by route — as it should:

| Route | Before | After |
|---|---|---|
| `openai` via **Codex OAuth**, `gpt-5.6*` | compacted at ~902K (over Codex capacity); display said 300K | capacity 372K → compacts at 352K |
| `openai` via **API key** | catalog 922K input, untouched | unchanged (the plugin only runs for `auth.type === "oauth"`) |
| **OpenRouter** / other resellers of the same model | their own catalog entry, untouched | unchanged |
| Codex OAuth, `gpt-5.3-codex` (272K input) | 252K | 252K — below the cap, untouched |
| Codex OAuth, `gpt-4o` (128K) | **raised to 263K** by #1926 | back to 91.6K |
| Codex OAuth, `gpt-image-*` (`context: 0`) | overflow handling **enabled** by #1926 | back to disabled |
| Any provider, no `max_context` set | — | byte-identical trigger to `main` |

Anyone who worked around #1926 with a `provider.openai.models.*.limit` override in their config keeps that override — it is merged before the plugin runs, and the plugin only clamps downward.

## Scope of a user-set budget

- **Per model, keyed `"<providerID>/<modelID>"`** — so `openai/gpt-5.6` (Codex) and `openrouter/openai/gpt-5.6` are independent. Wildcards allowed, longest pattern wins.
- **Global or per project** — `/context-limit` writes the user-global config; hand-editing a project `mimocode.json` scopes it to that project.
- **Affects when compaction fires, not what is sent.** The budget is not transmitted to the provider; it moves the auto-compaction / rebuild trigger, and with it the checkpoint thresholds and `preserveRecentBudget`, which are derived from the same number.
- **Never raises anything.** A budget ≥ the provider cap is a no-op and reports `source: "model"`; a budget that leaves no room for the reserves is ignored with a warning naming the model.
- **Applies to subagents in the session too**, since they resolve the same config and model.
- Not per session (deferred; see the spec's out-of-scope section) and not per agent.

## Testing

- `bun typecheck` — PASS
- `bun test test/plugin/codex.test.ts test/session/overflow.test.ts test/session/prune.test.ts test/session/checkpoint-thresholds.test.ts test/config` — 262 pass / 4 skip / 0 fail
- Full `bun test` before the review pass — 4359 pass / 4 fail, all reproduced at base or passing in isolation (`test/util/ssrf.test.ts` DNS fail-closed fails at base; `test/workflow/runtime.test.ts` has 2 fails at base vs 1 here; `test/session/checkpoint-rebuild-unify.test.ts` passes standalone in both trees)
- CLI: `MIMOCODE_CONFIG_CONTENT='{"compaction":{"max_context":{"openai/gpt-5*":"300K","openai/gpt-5.6":200000}}}' mimocode models openai` → `gpt-5.6` window 922K / budget 200K / compacts at 180K; `gpt-5.6-sol` budget 300K / 280K; `gpt-5.3-codex` (272K) shows no budget because 300K clamps away; `gpt-4o` and `o3` unchanged
- Live TUI (isolated `MIMOCODE_HOME`, `xiaomi/mimo-v2.5`): tier, custom `"50%"` and reset all persisted correctly and updated footer / sidebar / `/status` without a restart; a tier selection mid-stream left the config untouched, the same action once idle wrote it

## Notes for reviewers

- The footer percentage now reads **higher** than before for every model, because it is measured against the value that actually triggers compaction rather than the raw window. That is the intended correction.
- `packages/sdk/js` was **not** regenerated: `bun dev generate` emits a dangling `#/components/schemas/__schema0` ref from `ToolStateCompleted.providerOutput` and `@hey-api/openapi-ts` fails on it. Reproduced at base commit, so it is pre-existing; the new config field therefore needs one narrow cast at the TUI boundary.
- Design record: `docs/compose/spec/context-budget-control.md` (problem audit, rejected alternatives, and the four different numbers Codex calls "the context window").
- Optional follow-up, deliberately not in this PR: Codex applies the 272K billing line as a *default*. Matching that would need a provider-supplied default budget layer under the user's config in the resolution chain.


yanyihan-xiaomi:
## Second review pass + notes worth reading before merge

Pass 2 (fresh reviewer, full branch diff) confirmed all pass-1 criticals are genuinely fixed and independently verified two of the mechanisms — `jsonc-parser`'s `modify` writes a single property for a key containing `/` and `.` while preserving siblings and comments, and `session_status` deletes its entry on idle (`session/status.ts:73-79`) with the run-state scope finalizer as backstop, so the busy guard cannot false-negative during a turn nor lock a user out permanently.

It raised two new **important** findings, both fixed in 9cc844ea:

- `/context-limit` writes the user-global config, but every read is the *merged* config and project config wins per key (`config.ts:798-810`). A project `mimocode.json` could shadow the write while the toast reported success — the same class of lie an earlier finding was about. It now re-resolves after the write and reports the value still in force.
- The `/status` Context block described `local.model.current()` while dividing the *last assistant message's* tokens by it, so a mid-session model switch produced a wrong percentage. It now describes the model the tokens were actually spent on, matching the prompt footer.

### Known behaviour worth knowing about (not blockers)

**Your usable window depends on how you reach the model.** `limit.input` / `limit.context` are resolved per provider *and* per auth mode, so the same model name can have a different window on a ChatGPT/Codex subscription, a direct API key, and a reseller such as OpenRouter — and a catalog figure of 1M does not mean your route serves 1M. Concretely, `gpt-5.6-sol` is advertised at 1.05M context / 922K input, Codex's registry serves 372K, Codex's bundled metadata was lowered to 272K for billing reasons, and some plans are reported to be served 272K (openai/codex#33069). This PR only encodes the Codex-OAuth case; for any other route where you know the real ceiling is lower than the catalog says, set it yourself:

```jsonc
{
  "compaction": {
    "max_context": {
      "openrouter/openai/gpt-5.6": "272K"
    }
  }
}
```

`mimo models <provider>` prints the resolved window and the compaction point per model, so it is easy to check what MiMoCode thinks your route offers. `/context-limit` does the same from the TUI. Documented in the READMEs, `config.mdx`, and the `ln` skill reference (38acec67).

**372K can still be optimistic for some plans, and that is recoverable.** If the backend rejects an over-long prompt it surfaces as `ContextOverflowError` (`provider/error.ts:198`), which sets `needsOverflowHandling` and drives compaction instead of failing the turn (`processor.ts:768-772`). Cost is one wasted request until the user lowers `max_context`.

**`startsWith("gpt-")` is the only filter on the Codex clamp.** The largest behaviour change is `gpt-4.1*` (catalog 1,047,576, no `input`): its trigger moves from ~1.008M to 332K under Codex OAuth. The direction is safe — a clamp only lowers — and there is no evidence the Codex backend serves gpt-4.1 at all, but a narrower filter is a reasonable review request. `gpt-3.5-turbo`, `gpt-4o`, `gpt-realtime-2.1`, `gpt-5.3-codex-spark` and `gpt-image-*` are all below the cap and untouched.

**The "window" label is the prompt capacity, not the advertised context.** `hard = limit.input || limit.context`, so `/status` and `mimo models` print `window 922K` for a model marketed as 1.05M. That is the number that governs compaction; the difference is the output allowance.

Also deliberately out of scope: Codex applies its 272K billing line as a *default*. Matching that would need a provider-supplied default budget layer beneath the user's config in the resolution chain.

PR #1943, opened 2026-07-27
fix(plugin): skip non-function exports in getLegacyPlugins

Cipher208:
Fix: getLegacyPlugins throws TypeError on non-function exports, silently skipping plugins. Change throw→continue.


Cipher208:
Replaced by clean PR #1944 (bug fix only, no CI/CD changes)

PR #1958, opened 2026-07-28
fix(tui): discard superseded bootstrap writes and stop minting duplicate orchestrator roots

wqymi:
Follow-up to #1953 (predecessor, still open). Both PRs are self-contained and can merge in either order — this one is based on `main`, not on #1953's branch.

## The one root cause

#1953 established the principle: **check staleness AFTER the await, not before**. Two remaining defects live in the same window — the gap between `await sync.bootstrap()` returning and the writes it left in flight.

### Defect 1 — the non-blocking bootstrap writes were unguarded

`context/sync.tsx:780-802` launches ~11 requests as `void Promise.all([...])`, and on `main` every `.then` called `setStore` with no directory check at all. `dispose + switchDirectory + bootstrap` always spawns a second bootstrap through the `server.instance.disposed` handler (`sync.tsx:273`) whose requests were built from the **pre-switch** client, so a switch landing while these are in flight writes the old directory's `vcs` / `command` / `lsp` / `mcp` / `session` / … into a store that is supposed to describe the directory the client actually talks to.

Fix: capture the directory generation once and re-check it **after** each response resolves, via a single `guard()` helper rather than eleven copies of the same `if`. A superseded run also no longer flips `status` to `complete` — it must not unblock the UI on data it never wrote.

### Defect 2 — duplicate orchestrator roots (the observed symptom of that same window)

`app.tsx:540` read `sync.data.session` immediately after `await sync.bootstrap()`. But `session.list` only joins `blockingRequests` when `--continue` is set (`sync.tsx:726`), so **bootstrap resolves before the session list lands**. The `existing` lookup missed the real root and the `else` branch called `session.create({})`, minting a new one on each entry.

Fix: root resolution moves into `sync.session.resolveRoot()`, which refreshes from the server first, so resolve-or-create depends on data instead of on timing.

**Why not "promote `session.list` into `blockingRequests`"**: that would make every launch — not just orchestrator entry — wait on a 30-day, roots-only session list before the UI unblocks, a real startup regression on large histories. It would also leave correctness dependent on bootstrap's internal request classification, so a later refactor moving `session.list` around could silently reintroduce duplicate roots. Asking the server at the point of decision is both cheaper for everyone else and structurally immune to that.

**Why they must land together**: without Defect 1's guard, the stale bootstrap's session-list write can repopulate the store with the *launch* directory's roots after Defect 2 has resolved — which would point the Orchestrator at a session from another directory. The "strictly narrower race with no observed symptom" framing was wrong; Defect 2 is its symptom.

## Tests

`test/cli/tui/bootstrap-race.test.tsx`, driving the real contexts (`SDKProvider` + `ProjectProvider/ArgsProvider/ExitProvider/SyncProvider`) with a fetch double — the pattern #1953 introduced.

Both are proven regression tests by reverting only their src change:

- *a non-blocking bootstrap write that resolves after a directory switch is discarded* — with the post-resolve check removed:
  ```
  Expected: "branch-b"
  Received: "branch-a"
  ```
- *entering the orchestrator repeatedly resolves the one existing root instead of creating more* — with `resolveRoot` reverted to the old store read:
  ```
  error: expect(received).toEqual(expected)
    [
  -   "ses_orch",
  -   "ses_orch",
  -   "ses_orch",
  +   "ses_created_1",
  +   "ses_created_2",
  +   "ses_created_3",
    ]
  ```
  i.e. three entries, three roots — the reported symptom reproduced deterministically.

`test/cli/`: **266 pass / 0 fail** (688 expect, 37 files) vs pristine `60af8f1ff` baseline **264 pass / 0 fail** (683 expect, 36 files). `bun typecheck` exit 0.

## Live verification (`mimo/mimo-v2.5`, isolated dev home)

Five Orchestrator entries — three in-process (Tab in/out) plus two from fresh processes (cold store, the exact scenario the fix targets) — leave **exactly one** root:

```
ORCH ROOTS = 1   (all roots in db = 1)
    ses_057aa2b18ffeeg1zBgS7w6ZIfj | New session - 2026-07-28T10:47:05.192Z | created 18:47:05
```

created only at the first (legitimate) entry, when the server genuinely had none.

Honest caveat: a control build on pristine `main` in the same harness also produced only one root. The duplicate needs `session.list` to be slower than bootstrap's blocking group, and against a small local DB it is not — so the live run here is a regression check, not the discriminating proof. The discriminating proof is the unit test above, where the list latency is explicit.

## Deliberately left alone

- Bootstrap's **blocking** writes and `project.sync()` — that is #1953's scope. Consequence, still visible live on `main`: after entering Orchestrator the footer keeps showing the launch directory (the stale `instance.path`), which is exactly the blank-transcript bug #1953 fixes.
- `project.workspace.sync()` inside the non-blocking group (different context, #1953 territory).
- Pre-existing prettier drift in `app.tsx` / `sync.tsx` was left untouched so the diff stays reviewable; only the changed regions are formatted.

The `stale()` helper is named to match #1953's so a merge conflict resolves to simply keeping one definition.

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/app.tsx
+++ b/packages/opencode/src/cli/cmd/tui/app.tsx
@@ -537,23 +537,20 @@ function App(props: { onSnapshot?: () => Promise<string[]> }) {
           sdk.switchDirectory(dir)
           await sync.bootstrap()
         }
-        const existing = sync.data.session
-          .toSorted((a, b) => b.time.updated - a.time.updated)
-          .find((x) => x.parentID === undefined)?.id
-        if (existing) {
-          local.orchestrator.setSessionID(existing)
-          // A `-s` launch wanted to land IN the orchestrator session; a plain
-          // Tab-into-orchestrator from a stale launch-dir session wanted Home
-          // (the fresh-entry state). Either way navigate exactly once, AFTER
-          // bootstrap, so the switched view resolves directly to its target with
-          // no intermediate frame — the root now exists in orchestratorDir.
-          if (resumeIntoSession) route.navigate({ type: "session", sessionID: existing })
-          else if (switching) route.navigate({ type: "home" })
-        } else {
-          const res = await sdk.client.session.create({})
-          if (res.data?.id) local.orchestrator.setSessionID(res.data.id)
-          if (switching) route.navigate({ type: "home" })
-        }
+        // Authoritative resolve-or-create against the switched directory. Reading
+        // sync.data.session here raced bootstrap's NON-blocking session list —
+        // bootstrap resolves before the list lands, so the lookup missed the
+        // existing root and minted another one on every entry.
+        const root = await sync.session.resolveRoot()
+        if (root.id) local.orchestrator.setSessionID(root.id)
+        // A `-s` launch wanted to land IN the orchestrator session; a plain
+        // Tab-into-orchestrator from a stale launch-dir session wanted Home
+        // (the fresh-entry state). Either way navigate exactly once, AFTER
+        // bootstrap, so the switched view resolves directly to its target with
+        // no intermediate frame — the root now exists in orchestratorDir. A root
+        // we just created is empty, so resuming into it makes no sense: go Home.
+        if (root.id && !root.created && resumeIntoSession) route.navigate({ type: "session", sessionID: root.id })
+        else if (switching) route.navigate({ type: "home" })
       } catch (e) {
         toast.show({ message: `Failed to enter Orchestrator: ${e}`, variant: "error" })
       } finally {
--- a/packages/opencode/src/cli/cmd/tui/context/sync.tsx
+++ b/packages/opencode/src/cli/cmd/tui/context/sync.tsx
@@ -700,6 +700,21 @@ export const { use: useSync, provider: SyncProvider } = createSimpleContext({
         fullSyncedSessions.clear()
         syncedWorkspace = workspace
       }
+      // A bootstrap can outlive the directory it describes: `dispose +
+      // switchDirectory + bootstrap` ALSO re-fires bootstrap from the
+      // `server.instance.disposed` handler above, and that run built its requests
+      // from the PRE-switch client. Staleness therefore has to be re-checked AFTER
+      // each await rather than once before them — a switch landing while these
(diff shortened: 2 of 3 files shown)


wqymi:
Folded into **#1953** (`fix/tui-blank-transcript-dir-switch`) and closing here.

Rationale: both halves of this PR belong to #1953. Half (a) — the unguarded non-blocking bootstrap writes — is a review finding on #1953 itself: same file, same function, same defect family, and #1953 had already applied the after-the-await principle to the *blocking* writes. Half (b) — duplicate orchestrator roots — lives in the same unguarded window, and #1953's body had it listed as "separate bug, out of scope"; it isn't: without half (a)'s guard, the stale bootstrap's `session` list write can repopulate the store with the launch directory's roots *after* `resolveRoot()` resolved.

Commit `e096d9d0b` was cherry-picked onto #1953's branch (now `3f32482b9`). The expected `stale()` collision in `context/sync.tsx` **resolved to one definition**, exactly as this PR's naming intended: `const directory = sdk.directory` + `const stale = () => sdk.directory !== directory` are declared once (`sync.tsx:700`/`:722`) and now serve **both** write paths — #1953's blocking `if (stale()) return` guards *and* this PR's `guard()` helper wrapping all ~11 non-blocking writes. The two comment blocks were merged into one; nothing else was dropped. `app.tsx`'s `sync.session.resolveRoot()` call site, `resolveRoot()` itself, and `test/cli/tui/bootstrap-race.test.tsx` are all present on #1953's branch. No `prettier --write` was run on `app.tsx`/`sync.tsx`; the edit was surgical.

Post-fold on #1953: `test/cli/` = **268 pass / 0 fail** (38 files) vs pristine #1953 baseline **266 pass / 0 fail** (37 files) — both of this PR's tests survive. `bun typecheck` exit 0. Both revert probes were re-run after the merge and still bite: removing the post-resolve check in `guard()` → `Expected: "branch-b" / Received: "branch-a"`; removing `resolveRoot()`'s server refresh → `["ses_created_1", "ses_created_2", "ses_created_3"]` instead of three `ses_orch`.

Branch `fix/tui-bootstrap-stale-writes-dup-root` is left in place (not deleted).

PR #1963, opened 2026-07-28
fix(inbox): stop the drain from persisting an empty user text part (the Bedrock 400 producer)

wqymi:
## What this is

The **producer** behind the live Bedrock 400 `messages.<N>: user messages must have non-empty content`.

`64a128523` ("never PRODUCE an empty-content user message") guarded `createUserMessage`, and #1948 adds a pre-send backstop. Neither covers `Inbox.drain` — **the one user-message producer that does not go through `createUserMessage`**, so `hasSubstantiveContent` (`prompt.ts:4386`) never inspects it. This closes that producer.

#1948's `ensureNonEmptyContent` is deliberately left **intact** — defence in depth is wanted here.

## Root cause

- `renderInboxRow` (`src/inbox/render.ts:8`) used `content.text ?? "(no notification body)"`. `??` only catches `null`/`undefined`; a blank body is stored as `""` (`inbox.ts:162` `content: { text: input.content }`), and an `actor_notification` body is passed through **RAW** with no wrapper. So a blank notification rendered to `""`.
- `Inbox.drain` (`src/inbox/inbox.ts:268-277`) wrote one text part per row unconditionally, so that `""` was persisted as a user message whose only part is `{type:"text",text:""}` — `parts.length === 1`, which every `content.length === 0` check misses.
- `ai@6.0.168`'s `convertToLanguageModelMessage` user branch filters empty text parts out with **no backfill** (`.filter(p => p.type !== "text" || p.text !== "")`), so that message reaches the provider as `content: []` → the 400.
- The blank body is **reachable**: the `actor` tool's JSON path guards content with `z.string().min(1)` (`src/tool/actor.ts:454`) but the **shell** path (`src/tool/actor.ts:177-189`) did not, and `shell-wrap.ts:121` hands a parsed shell op straight to `execute` without re-validating it against `parameters`.

## Changes (three layers, outermost first)

1. **`src/tool/actor.ts`** — the shell `send` parser rejects a blank content, restoring parity with the JSON path's `.min(1)`.
2. **`src/inbox/render.ts`** — a `blankTo` helper substitutes the placeholder for any **blank** body, not just a missing one, for both row kinds. A blank notification is now visible to the model as `(no notification body)` rather than silently empty, so nothing is lost.
3. **`src/inbox/inbox.ts`** — the drain renders **before** writing anything and drops any row that renders blank; if every row renders blank it consumes the rows without writing a message at all. This is the structural invariant that keeps the fatal shape unreachable regardless of what a future row type renders.

## Tests

- **`test/inbox/drain-no-empty-part.test.ts`** (new, 7 tests): `renderInboxRow` never returns a blank string (empty / whitespace-only / missing body, plus verbatim passthrough for a real body); end-to-end, a blank `actor_notification` drained into a session yields exactly one **non-blank** synthetic part, and a blank mixed with a real one keeps both parts non-blank.
- **`test/tool/actor.shell.test.ts`** (+4): the shell `send` parser rejects `""` and whitespace-only content, and still accepts a short non-blank one.

### Each test proven to fail without its src change

| revert | result | observed failure |
|---|---|---|
| `render.ts` alone | 5 fail | the drain filter catches the blank row, so the notification is dropped: `expect(count).toBe(1)` → `Received: 0` |
| `render.ts` **and** `inbox.ts` (= pristine `main`) | 5 fail | the original defect shape is persisted: `expect(received).not.toBe("")` → `Expected: not ""` |
| `tool/actor.ts` alone | 3 fail | `expect(exit._tag).toBe("Failure")` → `Received: "Success"` |

## Verification

`bun typecheck` exit 0. Affected suites (`test/inbox/`, `test/tool/actor.shell.test.ts`, `test/tool/session-tool.test.ts`): **127 pass / 1 skip / 0 fail**.

Full `bun test` was run on a superset of this diff; its 15 failures are all pre-existing load-flaky families on the build box (the `prompt-effect` loop/cancel/busy suite, `Vcs` branch watchers, workflow timeouts) and none touch inbox, render, or the actor shell parser.

## Not in scope

Defect B — `recoverSessionArgs` silently rewriting a correct `send` into a `create` — ships separately as #1962.

### DIFF
--- a/packages/opencode/src/inbox/inbox.ts
+++ b/packages/opencode/src/inbox/inbox.ts
@@ -248,6 +248,44 @@ export const layer: Layer.Layer<
         return 0
       }
 
+      // Render BEFORE writing anything, and drop any row that renders blank.
+      // The drain is the one user-message producer that does NOT go through
+      // SessionPrompt.createUserMessage, so `hasSubstantiveContent` never sees
+      // it — a blank render would persist a user message whose only part is
+      // {type:"text",text:""}. The AI SDK's user branch filters empty text
+      // parts out with NO backfill, so that message reaches the provider as
+      // `content: []` and is rejected ("user messages must have non-empty
+      // content"). renderInboxRow already substitutes a placeholder for a
+      // blank body; this is the structural invariant that keeps the shape
+      // unreachable no matter what any future row type renders.
+      const rendered = rows.flatMap((row) => {
+        const text = renderInboxRow(row)
+        if (text.trim().length > 0) return [{ row, text }]
+        log.warn("inbox.drain: dropping row that rendered blank (would produce an empty user text part)", {
+          sessionID,
+          actorID,
+          rowID: row.id,
+          type: row.type,
+        })
+        return []
+      })
+
+      // Every row rendered blank: consume them (they carry no information and
+      // must not be re-drained forever) without writing a message at all. A
+      // zero-part user message would be skipped downstream anyway, so writing
+      // one is pure litter.
+      if (rendered.length === 0) {
+        yield* Effect.sync(() =>
+          Database.use((db) =>
+            db
+              .delete(InboxTable)
+              .where(inArray(InboxTable.id, rows.map((r) => r.id)))
+              .run(),
+          ),
+        )
+        return 0
+      }
+
       // Non-transactional crash window: updateMessage + updatePart commit
       // before the inbox DELETE. A crash between them re-renders the same
       // rows on next drain — LLM sees duplicated notifications. Tolerable;
@@ -265,14 +303,14 @@ export const layer: Layer.Layer<
         agent: seed.agent,
         model: seed.model,
       })
-      for (const row of rows) {
+      for (const entry of rendered) {
         yield* sessions.updatePart({
           id: PartID.ascending(),
           messageID: msgID,
           sessionID,
           type: "text" as const,
           synthetic: true,
-          text: renderInboxRow(row),
+          text: entry.text,
         })
       }
       yield* Effect.sync(() =>
@@ -284,7 +322,7 @@ export const layer: Layer.Layer<
         ),
       )
 
-      return rows.length
+      return rendered.length
     })
 
     const impl = Service.of({ send, drain })
--- a/packages/opencode/src/inbox/render.ts
+++ b/packages/opencode/src/inbox/render.ts
@@ -1,11 +1,21 @@
 import type { InboxRow } from "./inbox.sql"
 
+// A blank body must fall back to the placeholder, not just a missing one.
+// `?? placeholder` only catches null/undefined, but a blank body is stored as
+// "" (or whitespace) — and for actor_notification the body is passed through
+// RAW, so "" would become a user text part with text:"". The AI SDK's user
+// branch filters empty text parts out with no backfill, leaving `content: []`
+// and a provider 400 ("user messages must have non-empty content").
+function blankTo(text: string | undefined, placeholder: string) {
+  return text !== undefined && text.trim().length > 0 ? text : placeholder
+}
+
 export function renderInboxRow(row: InboxRow): string {
   if (row.type === "actor_notification") {
     // Pre-rendered notification text — sender produced the full
     // <actor-notification>...</actor-notification> wrapper.
     const content = row.content as { text?: string }
-    return content.text ?? "(no notification body)"
+    return blankTo(content.text, "(no notification body)")
   }
   // Default: type === "text" or unknown — wrap as <inbox> element so
   // the LLM can route by sender; the wrapper format mirrors the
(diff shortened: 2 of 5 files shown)


wqymi:
Folded into #1948 and closing this in favour of it. The user's rule is that one defect family is one PR, and this PR and #1948 were editing the same two files (`src/inbox/render.ts`, `src/tool/actor.ts`) with two different versions of the same fix.

**Nothing from here was dropped — this PR's version won every overlap:**

- `src/inbox/render.ts` — kept **this PR's** `blankTo()` helper over #1948's bare `??` → `||`. It catches whitespace-only bodies, not just `""`, and it is applied to the `<inbox>` branch as well.
- `src/inbox/inbox.ts` — kept **this PR's** drain-renders-before-writing invariant in full, including the "every row rendered blank ⇒ consume without writing a message" arm. This is the structural guarantee and #1948 had no equivalent.
- `src/tool/actor.ts` — kept **this PR's** `.trim()` condition over #1948's `=== ""`.
- `test/inbox/drain-no-empty-part.test.ts` — moved over unchanged.
- `test/tool/actor.shell.test.ts` — kept **this PR's** parametrised block (3 blank cases + the not-over-broad case). #1948's single case used the identical script and assertion, so it is the 2nd loop iteration; no assertion was lost.

Verified in the fold: reverting `src/inbox/inbox.ts` + `src/inbox/render.ts` to `main` on the merged branch produces **10 failures split across both sides' test files** (5 from this PR's `drain-no-empty-part.test.ts`, 5 from #1948's `empty-notification-part.test.ts`), so the fold neutralised nothing.

**Two corrections to the framing carried over, worth recording here:**

1. This PR's `src/tool/actor.ts` comment ("shell-wrap hands a parsed shell op straight to execute WITHOUT re-validating it against `parameters`") is **false**, and #1948 now carries a corrected version. `shell-wrap.ts:121` calls `def.execute(parsed, ctx)` on the def from `Tool.init`, which is `wrap()`-decorated, and `wrap()` runs `parameters.parse(args)` *inside* `execute`. Driving the real composition `shellWrap(Tool.init(actor)).execute({ script: 'actor send <id> ""' })` **with this PR's guard removed** still enqueues nothing and reports `Too small: expected string to have >=1 characters → at operation.content`. A revert probe on `parseActorScript` cannot settle this — asserting that the parser fails and then removing the parser's guard is circular.

2. Consequently the render/drain work is described in #1948 as closing a defence gap that no observed transcript exercised (0 user text parts with `text = ''` in a 5.6 GB production DB), not as fixing an observed producer. The **actual** producer turned out to be elsewhere and is now the headline of #1948: `ensureTrailingUserMessage` appended `{ role: "user", content: "Continue." }` as a bare **string** into a `LanguageModelV3Prompt`, and `@ai-sdk/anthropic` iterates a string character-by-character with no `default` case, shipping `{"role":"user","content":[]}`.

The guard added here is kept on its own merits: it turns a generic zod dump into one specific, teachable message, and `.trim()` rejects whitespace-only bodies which `min(1)` accepts.

Branch `fix/session-msgshape-dual` is left in place (not deleted).

wqymi:
Closing in favour of **#1948**, which now carries this entire defect family in one PR (the user's rule: one defect family, one PR — not scattered across three).

**Nothing here is lost. #1963's version won every overlap:**

- `src/inbox/render.ts` — #1963's `blankTo()` helper replaces #1948's bare `??` → `||`. Strictly stronger: it catches whitespace-only bodies, and it covers the `<inbox>` branch too.
- `src/inbox/inbox.ts:268-277` — the drain-renders-before-writing invariant, taken wholesale. This is the structural guarantee #1948 did not have, and it is the right layer: the drain is the one user-message producer that bypasses `createUserMessage`/`hasSubstantiveContent` (`prompt.ts:2164`, `:4386`), which #1963 correctly identified.
- `src/tool/actor.ts` — #1963's `.trim()` condition kept over #1948's `=== ""`.
- `test/inbox/drain-no-empty-part.test.ts` — added verbatim.
- `test/tool/actor.shell.test.ts` — #1963's parametrised 3-blank-cases + not-over-broad case kept; #1948's single case was byte-for-byte the 2nd iteration of that loop with the same assertion, so no assertion was dropped.

Verified the fold did not neutralise anything: reverting `inbox.ts` + `render.ts` to `origin/main` on the merged branch fails **10** tests, split 5/5 across *both* folded test files. Folded branch runs 535 pass / 0 fail / 24 files vs this branch's pristine 507 pass / 0 fail / 21 files, identical command. `bun typecheck` exit 0.

Two corrections applied while folding, both worth reading:

1. The shell-boundary guard's comment (carried by both PRs) asserted that a shell-parsed op is "NEVER re-validated against `parameters`". That is **false** — `shell-wrap.ts:121` calls `def.execute(parsed)` on the def from `Tool.init`, which is `wrap()`-decorated, and `wrap()` runs `parameters.parse(args)` inside `execute`. Driving the real composition with the guard removed still enqueues nothing and reports `Too small: expected string to have >=1 characters → at operation.content`. The guard is kept for message quality and for the whitespace case, not for reachability. Comment corrected in #1948.
2. The real producer of `content: []` turned out to be elsewhere and is now #1948's headline: `ensureTrailingUserMessage` appended `{ role: "user", content: "Continue." }` as a **bare string** into a `LanguageModelV3Prompt`, and `@ai-sdk/anthropic` iterates a string character-by-character through a `switch (part.type)` with no `default`, shipping `content: []`. Wire-proven against a captured outbound HTTP body.

The branch `fix/session-msgshape-dual` is left in place (not deleted).

PR #1978, opened 2026-07-29
fix(tool): share the realpath-comparing trust-boundary test instead of leaving it in one file

wqymi:
## The defect

`AppFileSystem.contains` (`packages/shared/src/filesystem.ts:233`) is pure lexical path arithmetic:

```ts
export function contains(parent: string, child: string) {
  return !relative(parent, child).startsWith("..")
}
```

No realpath, no symlink normalisation. Several callers nevertheless use it as a **trust boundary**, and one of them gates a deadlock. `tool/external-directory.ts` trusts the app-managed worktree base; the comment above that line says what the early return prevents — *every in-worktree write hits `external_directory:ask`, and a background/isolated child has no interactive replier, so the ask hangs on a never-resolved `Deferred` and the child deadlocks*. A false negative there re-introduces that deadlock.

**The hazard does not require a symlink to exist.** `Instance.provide` stores a **realpath-resolved** directory (`instance.ts:121`, via `AppFileSystem.resolve`), while `Global.Path.data` keeps whatever spelling it was configured with. It is enough that one side has been resolved and the other has not — the two sides merely need different degrees of normalisation.

**The repo had already fixed this in exactly one place.** `tool/isolated-git-guard.ts` carried a local two-step test, and its own doc comment named the sibling call site (`src/tool/external-directory.ts`) that still had the hole — so whoever wrote it knew the other callers shared the bug and fixed only their own.

## What this changes

Promotes that two-step test into one shared predicate, `AppFileSystem.withinTrustedRoot(root, child)`, and applies it at the boundary call sites.

`contains` itself is **unchanged**. Widening it would silently alter every caller's semantics, including those that genuinely want lexical answers. The pre-existing test that pins its lexical behaviour (`packages/shared/test/filesystem/filesystem.test.ts`) still passes untouched, which is itself the evidence.

### Where the predicate lives, and why

`packages/shared/src/filesystem.ts`, directly below `contains`.

The concern with putting it in `shared` is dragging a node dependency into a package other consumers use. **Checked before choosing — that concern is already moot for this exact file:** `shared/src/filesystem.ts:3` already imports `realpathSync` from `"fs"` and uses it at `:193` (`normalizePath`) and `:211` (`resolve`). `@effect/platform-node` is already a declared dependency. So this adds no new dependency edge, and it puts the two predicates side by side where a reader choosing between them sees both. `packages/opencode` already imports `AppFileSystem` at every affected site, `isolated-git-guard.ts` included.

### Caller table — all 11 sites of `AppFileSystem.contains`

Inventory re-derived on `origin/main` @ `b81670870`; matches, 11 sites.

| site | semantics | action |
|---|---|---|
| `tool/isolated-git-guard.ts:266` | trust boundary | **now delegates** to the shared predicate; local `real()` copy deleted so the two cannot drift |
| `tool/external-directory.ts:67` (`<data>/worktree`) | trust boundary, **gates the deadlock** | **promoted** |
| `project/instance.ts:170`, `:174` | in-project test, first gate of `external_directory` | **promoted** |
| `tool/external-directory.ts:46` (`<data>/memory`) | trust boundary, **deferral** | left lexical — see below |
| `tool/external-directory.ts:168` (`<data>/memory`) | trust boundary, **deferral** | left lexical — see below |
| `tool/apply_patch.ts:198`, `:214` (`<data>/memory`) | trust boundary, **deferral** | left lexical — see below |
| `lsp/lsp.ts:236`, `:237` | out-of-project → skip LSP | left lexical, deliberately |

### Why the four `<data>/memory` sites were deliberately NOT promoted

This is the one place the brief's plan and the code disagreed, and the code won.

Those four are not terminal trusts — they are **deferrals**. The comment at `external-directory.ts` says so: the memory tree *"has its own finer authority (memory-path-guard)"*. Their soundness depends on the delegate actually governing what is waved through. **The delegate is still lexical:**

- `tool/memory-path-guard.ts:168` — `assertMemoryWriteAllowed` does `if (!target.startsWith(normalizedRoot)) return`, so a memory path spelled through a symlinked data root is governed by **nothing**.
- `tool/memory-path-guard.ts:35-36` — `assertAgentWriteSandbox` resolves both sides with `path.resolve`, which does not follow symlinks.

So widening only the deferral would **drop the prompt for exactly the paths the delegate never inspects** — a net loss of oversight, not a fix. Demonstrated, not theorised: an aliased memory write under `checkpoint-writer` fails closed in the guard with

```
error: Agent 'checkpoint-writer' may only write under the memory tree.
  memory: /var/folders/0s/fm84ntp94712pvplln0lf9940000gn/T/mimocode-test-data-28641/share/mimocode/memory
You attempted: /private/var/folders/0s/fm84ntp94712pvplln0lf9940000gn/T/applypatch-datalink-6RzBQo/datalink/memory/sessions/ses_test/aliased-checkpoint.md
      at assertAgentWriteSandbox (src/tool/memory-path-guard.ts:39:15)
```

— the `/var/...` root against the `/private/var/...` target is this very bug, live, in a third file. The memory sites and `memory-path-guard.ts` must be promoted **together**; the reason is recorded at each site, and a test pins the current answer so a future change has to confront it rather than silently flip it.

### `lsp.ts`: left lexical, on purpose

Not a trust boundary — a false negative only skips LSP enrichment for one file — and `getClients` runs on every LSP touch. Paying realpath syscalls per out-of-project file is the wrong trade. Recorded at the site, including which line to change if LSP ever stops attaching under a symlinked prefix.

### The fallback is stronger than the copy it replaces

The guard's `real()` was `catch -> path.resolve`. `path.resolve` performs **no** symlink resolution, so a **not-yet-existing** path keeps its unresolved spelling and the comparison still misses — and *"may I create this file"* is the commonest boundary question on the write path. This was caught by the end-to-end test failing **with the fix applied**, not by inspection.

So `realpathBestEffort` canonicalizes the deepest ancestor that **does** exist and re-appends the remainder — the technique already proven in-tree at `tool/tool-script.ts:250`. The load-bearing property is preserved and strengthened: it never throws, and a path that does not exist yet is still comparable — now also when it is named *through* a symlink.

## Cost: worst-case syscall rate on the permission path

Ordering preserved: **lexical test first, realpath only on the negative path.**

`realpathBestEffort` costs 1 syscall, plus 1 per missing trailing component. One `withinTrustedRoot` miss therefore costs 2 resolutions; for the realistic shape (a new file under an existing directory) that is ~3 syscalls.

Per `assertExternalDirectoryEffect` call:

- **In-project write, matching spelling — the hot case: 0 extra syscalls.** `Instance.containsPath`'s lexical test hits and returns immediately.
- **Project under a symlinked prefix (the case this fixes) — worst *sustained* rate: ~3 `realpathSync` per file-tool call.** `containsPath` misses lexically, resolves, hits. Negligible beside the file I/O the tool is about to perform.
- **Genuinely external path — worst case ~9.** Up to three `withinTrustedRoot` evaluations (`directory`, `worktree`, `<data>/worktree`) all miss. This is the call that then raises an interactive `external_directory:ask`, i.e. a human round-trip measured in seconds. The memory branch stays lexical and adds 0.

## Verification

Baseline and post-change runs are the **same commands at the same sha**, `b8167087088ad361740d6e28fb35faae49afdf78`:

| command | baseline | after |
|---|---|---|
| `packages/shared` `bun test test/filesystem/filesystem.test.ts` | 24 pass / 0 fail / 35 expects | **32 pass / 0 fail / 52 expects** |
| `packages/opencode` `bun test test/tool/{external-directory,isolated-git-guard,apply_patch}.test.ts` | 133 pass / 0 fail / 178 expects | **139 pass / 0 fail / 189 expects** |
| `bun typecheck` `packages/shared` | exit 0 | **exit 0** |
| `bun typecheck` `packages/opencode` | exit 0 | **exit 0** |
| `bun lint` (oxlint) | 3930 warnings / 0 errors | **3930 warnings / 0 errors** (identical; 24 file-specific warnings both before and after) |

Repo-wide turbo typecheck via the pre-push hook: **12 successful, 12 total.**

### The fixture is a real symlink, not a mock

`<tmp>/link -> <tmp>/real`, so one file can be named through either spelling; plus a symlink pointing at the actual data root, so the two `<data>`-rooted gates are exercised with one side canonical and the other not. Nothing stubs `contains`.

### Revert probes — each quoted verbatim

**P1 — `project/instance.ts` back to lexical `contains`:**
```
227 |     expect(requests.length).toBe(0)
                                  ^

Expected: 0
Received: 1

(fail) tool.assertExternalDirectory > does NOT ask for an in-project path named through a symlinked project prefix [2.34ms]
 13 pass  1 fail
```

**P2 — the `<data>/worktree` gate back to lexical `contains` (isolated):**
```
283 |     expect(requests.length).toBe(0)
                                  ^

Expected: 0
Received: 1

(fail) tool.assertExternalDirectory > does NOT ask for a worktree path named through an aliased data root [1.09ms]
 13 pass  1 fail
```

**P3 — fallback replaced by the old `catch -> path.resolve` semantics:**
```
(fail) AppFileSystem > withinTrustedRoot > a not-yet-existing child named THROUGH the symlink is still inside [1.10ms]

415 |       expect(AppFileSystem.withinTrustedRoot(real, deep)).toBe(true)
                                                                ^

Expected: true
Received: false

(fail) AppFileSystem > withinTrustedRoot > a deep chain of missing components resolves without throwing [0.54ms]
 30 pass  2 fail
```

**P4 — `isolated-git-guard.ts` keeps its own local two-step copy instead of delegating:**
```
199 |     expect(isIsolatedWorktree(path.join(linkBase, "p_abc", "child-not-created-yet"), realBase)).toBe(true)
                                                                                                      ^

Expected: true
Received: false

(fail) isolated-git-guard / isolation signal > a symlinked worktree base still recognises its own children [3.68ms]
 92 pass  1 fail
```

P4 confirms the delegation is a behaviour improvement, not a cosmetic refactor. `lsp.ts` is comment-only and has no probe by construction.

## Comments corrected

Stale comments that would have described behaviour the code no longer has:

1. **`isolated-git-guard.ts:256-262`** — said it *"Mirrors the trusted-root test already used by `src/tool/external-directory.ts`, but additionally compares realpaths"*. Both halves were about to become wrong: it no longer holds a local special case, and the sibling now applies the same predicate. Rewritten to state the delegation and the anti-drift reason.
2. **`external-directory.ts`, worktree branch** — extended to say why the predicate must be realpath-comparing, and that this trust is *terminal* (nothing downstream re-checks it), which is what distinguishes it from the memory branch.
3. **`external-directory.ts`, memory branch** — new note recording that the lexical test is a deliberate choice and naming the coupling to `memory-path-guard`.
4. **`askEditUnlessMemory` doc comment** — previously claimed only to mirror the memory deferral; now also records that it mirrors its *lexical* choice, and why.
5. **`apply_patch.ts` permission filter** — same note, so the two sites are not read as an oversight.
6. **`instance.ts` `containsPath` doc comment** — records that it is the first gate of `external_directory` and why `Instance.provide`'s realpath resolution forces the boundary predicate.
7. **`lsp.ts` `getClients`** — new comment recording the deliberate non-change and its cost reasoning.

## Also found, not fixed here

- **A second, duplicated lexical `contains`** at `packages/opencode/src/util/filesystem.ts:164`, byte-identical to the shared one, with 6 callers of its own — including `server/routes/instance/middleware.ts:37` and `server/routes/instance/httpapi/server.ts:106`, which gate a **403 directory-access decision** and compare an already-`Filesystem.resolve`d `cwd` (realpath) against a request-supplied `directory` (not). Same asymmetry, in HTTP authorization. Not touched here: changing an auth path needs its own reasoning and tests.
- **`memory-path-guard.ts:35-36` and `:168`** — the delegate discussed above.
- **`workflow/workspace.ts:21`** — a sandbox jail whose comment already documents the limitation as deferred-by-design. Worth noting `withinTrustedRoot` would **not** fix it: its hazard is a symlink *inside* the root pointing *out* (a false positive), and the predicate short-circuits on the lexical `true`. Applying it there would look like a fix while changing nothing.
- **`tool/tool-script.ts:250`** — a third independent reimplementation of the realpath dance, which is the evidence that this pattern was being reinvented per-site.

### DIFF
--- a/packages/opencode/src/lsp/lsp.ts
+++ b/packages/opencode/src/lsp/lsp.ts
@@ -226,6 +226,12 @@ export const layer = Layer.effect(
 
     const getClients = Effect.fnUntraced(function* (file: string) {
       const ctx = yield* InstanceState.context
+      // Deliberately the LEXICAL `contains`, not the realpath-comparing
+      // `withinTrustedRoot` used by the permission gates. This is not a trust
+      // boundary: a false negative only skips LSP enrichment for one file, and
+      // getClients runs on every LSP touch, so paying two realpath syscalls per
+      // out-of-project file would be the wrong trade. If LSP silently stops
+      // attaching on a symlinked project prefix, this is the line to change.
       if (
         !AppFileSystem.contains(ctx.directory, file) &&
         (ctx.worktree === "/" || !AppFileSystem.contains(ctx.worktree, file))
--- a/packages/opencode/src/project/instance.ts
+++ b/packages/opencode/src/project/instance.ts
@@ -157,14 +157,21 @@ export const Instance = {
    * Check if a path is within the project boundary.
    * Returns true if path is inside Instance.directory OR Instance.worktree.
    * Paths within the worktree but outside the working directory should not trigger external_directory permission.
+   *
+   * This is the first gate of the external_directory permission check, so it uses
+   * the trust-boundary predicate rather than a lexical prefix test: `Instance.provide`
+   * stores a REALPATH-RESOLVED directory, while a filepath arriving from a tool call
+   * carries whatever normalisation the caller gave it. A lexical-only comparison of
+   * two differently-normalised paths returns a false negative and sends an in-project
+   * write to the ask.
    */
   containsPath(filepath: string, ctx?: InstanceContext) {
     const instance = ctx ?? Instance
-    if (AppFileSystem.contains(instance.directory, filepath)) return true
+    if (AppFileSystem.withinTrustedRoot(instance.directory, filepath)) return true
     // Non-git projects set worktree to "/" which would match ANY absolute path.
     // Skip worktree check in this case to preserve external_directory permissions.
     if (instance.worktree === "/") return false
-    return AppFileSystem.contains(instance.worktree, filepath)
+    return AppFileSystem.withinTrustedRoot(instance.worktree, filepath)
   },
   /**
    * Captures the current instance ALS context and returns a wrapper that
--- a/packages/opencode/src/tool/apply_patch.ts
+++ b/packages/opencode/src/tool/apply_patch.ts
@@ -189,6 +189,11 @@ export const ApplyPatchTool = Tool.define(
       const title = files.length === 1 ? files[0]!.relativePath : `${files.length} files`
 
       // Check permissions if needed
+      // Deliberately the lexical `contains`: this mirrors the memory-region deferral
+      // in external-directory.ts, whose delegate (memory-path-guard) is still lexical.
+      // Widening only this filter would drop the `edit` ask for memory paths spelled
+      // through a symlinked data root — paths assertMemoryWriteAllowed skips entirely.
+      // See the note at the memory branch of assertExternalDirectoryEffect.
       const permissionChanges = fileChanges.filter(
         (change) => !AppFileSystem.contains(path.join(Global.Path.data, "memory"), change.movePath ?? change.filePath),
       )
--- a/packages/opencode/src/tool/external-directory.ts
+++ b/packages/opencode/src/tool/external-directory.ts
@@ -34,6 +34,15 @@ export const assertExternalDirectoryEffect = Effect.fn("Tool.assertExternalDirec
   // is redundant and, in headless run mode (no permission replier), deadlocks on a
   // never-resolved Deferred. memory-path-guard allows a task-bound subagent its own
   // tasks/<taskId>/*.md and rejects cross-task / wrong-agent writes.
+  //
+  // DELIBERATELY the lexical `contains`, not withinTrustedRoot — unlike the worktree
+  // base below, this early return is a DEFERRAL, and its soundness depends on the
+  // delegate actually governing what is waved through here. memory-path-guard is
+  // still lexical: assertMemoryWriteAllowed returns early on `!target.startsWith(root)`,
+  // so a memory path spelled through a symlinked data root is governed by NOTHING.
+  // Widening this test alone would therefore drop the prompt for exactly the paths
+  // the guard does not inspect. These two must be promoted together; until then the
+  // conservative answer (ask) is the safe one.
   if (AppFileSystem.contains(path.join(Global.Path.data, "memory"), full)) return
 
   // Orchestrator-created worktrees live under <data>/worktree/<projectID>/<name>.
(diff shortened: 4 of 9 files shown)


wqymi:

PR #1984, opened 2026-07-30
fix(tui): reconcile the three store writes that mean "replace", not "merge"

wqymi:
## What

Three TUI store writes mean **replace**, but Solid's store setter **merges** plain objects into the existing node — `mergeStoreNode` only writes `Object.keys(next)`, so an omitted key silently inherits its old value. Each of these three is authoritative for the whole object, so each was leaking a field across a transition that logically clears it.

Same `reconcile()` remedy and the same named-helper shape as the existing `nextSessionStatus` in `context/sync.tsx` (the #1752 fix), so this is a consistency pass over the remaining offenders rather than a new pattern.

## The three leaks

**`ui/toast.tsx` — `currentToast`**
`title` is `z.string().optional()` and `toast.error(err)` omits it *structurally* (`{ variant, message }`). `app.tsx:1339` shows a titled success toast with `duration: 10000`, so any error raised in that 10-second window landed on a non-null previous toast and rendered under the unrelated "updated" headline — attributing a failure to work that succeeded.

**`component/prompt/frecency.tsx` — `data`**
Both whole-map writes are **prunes**: they hand over the survivors and mean the rest to be gone. Merging meant the dropped keys stayed in the store while the `.jsonl` was rewritten without them, so `MAX_FRECENCY_ENTRIES` never took effect in memory — `store.data` grew unbounded, the over-cap branch re-fired on every open, and dropped paths kept scoring in the file picker.

**`context/sync.tsx` — `workflow[runID]`**
`workflow.started` mints a *fresh* run row, but a resume reuses the **same** `runID`: `WorkflowRuntime.resume` hands `input.runID` to `launch` (`workflow/runtime.ts:1577`), which republishes `WorkflowStarted` (`:712`). The "fresh" row therefore inherited the dead attempt's `currentPhase` and `error`, so `/workflows` showed a *running* run stamped with the phase it died in and the error it had just retried past. Note `loadWorkflows` already wrote this same node with `reconcile(run)` — only the event handler disagreed, and the two writers are now pinned together.

## Deliberately not changed

Targeted single-field writes stay merges, and are pinned as such so a blanket reconcile sweep cannot break them:
`setStore("data", absolutePath, newEntry)`, `setStore("workflow", runID, "currentPhase", …)`, `setStore("workflow", runID, "status", …)`, and the `setStore("currentToast", null)` expiry (null is not an object, so it replaces already).

`component/prompt/index.tsx`'s `setStore("prompt", { input: "", parts: [] })` was examined and **left alone**. It does structurally omit `PromptInfo.mode?`, which history navigation writes onto the node (`setStore("prompt", item)`), so the field really does survive the clear — but the single consumer is `history.append({ ...store.prompt, mode: currentMode })`, where `mode` is re-specified *after* the spread, and `store.prompt.mode` is read nowhere else in the tree. The leak is inert; changing it would be churn without a behaviour delta.

## Tests

`test/cli/tui/{toast,frecency,workflow-run}-store.test.ts`, following the existing `session-status-store.test.ts` idiom.

The real stores live inside provider factories (`init()` is module-private and calls `useLanguage()`; `SyncProvider` needs sdk/project contexts plus a bootstrap) and there is **no Solid render harness for this route**. So each file pairs:

1. **behavioural** tests driving a store wired exactly as production wires it — these pin the helper's semantics; and
2. a **source-level, explicitly non-behavioural** assertion on the production write site — these pin the *wiring*, which (1) cannot see.

Both halves are independently revert-probed: de-reconciling the helpers fails 6 behavioural tests; leaving the helpers correct but unwiring the three call sites fails exactly the 3 source-level tests. The in-file comments label which is which rather than implying behavioural coverage of the wiring.

## Verification

- `bun test test/cli/tui`: **152 pass → 171 pass** (+19). The 1 pre-existing fail / 1 error is unchanged — `thread.test.ts` aborted by a cross-file module-load error in `src/workflow/builtin/deep-research.js`, reproducible on unmodified `b8167087088ad361740d6e28fb35faae49afdf78` and unrelated to this change.
- `bun typecheck --force`: 12 successful, `Cached: 0 cached`.
- `bunx oxlint` on the six touched files: 0 errors; 14 warnings, all pre-existing and byte-identical to the same lint run on the pristine sources.

## Relationship to #1978

#1978 (`fix/trust-boundary-realpath`) owns the trusted-root/symlink work. `src/tool/external-directory.ts` is **untouched here** — verified identical to `origin/main` on the pushed branch. See the review note below for why the overlapping half of the stale branch this was recovered from was discarded rather than merged.

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/frecency.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/frecency.tsx
@@ -2,7 +2,7 @@ import path from "path"
 import { Global } from "@/global"
 import { Filesystem } from "@/util"
 import { onMount } from "solid-js"
-import { createStore } from "solid-js/store"
+import { createStore, reconcile } from "solid-js/store"
 import { createSimpleContext } from "../../context/helper"
 import { appendFile, writeFile } from "fs/promises"
 
@@ -11,6 +11,21 @@ function calculateFrecency(entry?: { frequency: number; lastOpen: number }): num
   const daysSince = (Date.now() - entry.lastOpen) / 86400000 // ms per day
   const weight = 1 / (1 + daysSince)
   return entry.frequency * weight
+}
+
+/**
+ * Both writes of the whole `data` map are PRUNES: they hand over the surviving
+ * entries and mean the rest to be gone.
+ *
+ * Solid's store setter merges plain objects into the existing node
+ * (`mergeStoreNode` only writes `Object.keys(next)`), so the pruned keys used to
+ * survive in the store while the on-disk file was rewritten without them. The
+ * MAX_FRECENCY_ENTRIES cap therefore never took effect in memory: `store.data`
+ * grew without bound, every subsequent open re-ran the prune, and dropped paths
+ * kept scoring in the file picker.
+ */
+export function nextFrecencyData(data: Record<string, { frequency: number; lastOpen: number }>) {
+  return reconcile(data)
 }
 
 const MAX_FRECENCY_ENTRIES = 1000
@@ -47,8 +62,10 @@ export const { use: useFrecency, provider: FrecencyProvider } = createSimpleCont
 
       setStore(
         "data",
-        Object.fromEntries(
-          sorted.map((entry) => [entry.path, { frequency: entry.frequency, lastOpen: entry.lastOpen }]),
+        nextFrecencyData(
+          Object.fromEntries(
+            sorted.map((entry) => [entry.path, { frequency: entry.frequency, lastOpen: entry.lastOpen }]),
+          ),
         ),
       )
 
@@ -75,7 +92,7 @@ export const { use: useFrecency, provider: FrecencyProvider } = createSimpleCont
         const sorted = Object.entries(store.data)
           .sort(([, a], [, b]) => b.lastOpen - a.lastOpen)
           .slice(0, MAX_FRECENCY_ENTRIES)
-        setStore("data", Object.fromEntries(sorted))
+        setStore("data", nextFrecencyData(Object.fromEntries(sorted)))
         const content = sorted.map(([path, entry]) => JSON.stringify({ path, ...entry })).join("\n") + "\n"
         writeFile(frecencyPath, content).catch(() => {})
       }
--- a/packages/opencode/src/cli/cmd/tui/context/sync.tsx
+++ b/packages/opencode/src/cli/cmd/tui/context/sync.tsx
@@ -65,6 +65,23 @@ export type WorkflowRun = {
   updatedAt?: number
 }
 
+/**
+ * A `workflow.started` event mints a FRESH run row, so it is authoritative for
+ * the whole object — including the optional fields it omits.
+ *
+ * Solid's store setter merges plain objects into the existing node
+ * (`mergeStoreNode` only writes `Object.keys(next)`), and a resume reuses the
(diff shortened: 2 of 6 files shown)


wqymi:
### Provenance / why the overlapping half was discarded

This work was recovered from a pushed-but-never-opened branch, `refs/heads/fix/setstore-merge-audit-and-symlink` @ `c12bda4fa` (2026-07-28, 8 files, +437/−18). That branch tangled **two** concerns, which is likely why it stalled: a `setStore` merge-semantics half (landed here) and a trusted-root/symlink half that overlaps #1978 in `src/tool/external-directory.ts` + its test. Only the first half is in this PR; `external-directory.ts` on this branch is byte-identical to `origin/main`.

Both halves were diffed against the **same** base (`b8167087088ad361740d6e28fb35faae49afdf78`) and compared behaviourally, not textually.

**Verdict on the trusted-root half: fully superseded by #1978, and its one non-overlapping part is a regression — discarded, not merged.**

- **Algorithm: equivalent, so no residue.** The stale branch's private `real()` already used a deepest-existing-ancestor walk, *not* the weaker `catch → path.resolve` (which does no symlink resolution and would miss the dominant "may I create this file" case). #1978's `AppFileSystem.realpathBestEffort` is the same technique, and #1978 additionally promotes it to a shared predicate reused by `isolated-git-guard`, `project/instance.ts` and the `<data>/worktree` gate, instead of leaving a fourth private copy in one file. The stale branch's own comment even anticipated this ("once both land, that one should delegate here").
- **Tests: #1978 is a strict superset.** Both use real `fs.symlink` fixtures, not stubs, and both pin the lexical asymmetry so the fixture cannot pass for the wrong reason. #1978 additionally covers both symlink directions, deep missing-component chains, neither-side-existing, and a name-resembling-but-foreign root in `packages/shared/test/filesystem/filesystem.test.ts`. The only case the stale tests exercised uniquely (leaf *exists*, so `realpathSync` succeeds on the first try) is covered there by "root named through a symlink still contains a child named through the real path".
- **Its one unique change is the hole #1978 explicitly refused to open.** The stale branch also widened the two `<data>/memory` sites, and did **not** touch `tool/memory-path-guard.ts`, whose `assertMemoryWriteAllowed` still returns early on `!target.startsWith(normalizedRoot)` — lexical. So for a memory path spelled through a symlinked data root: `assertExternalDirectoryEffect` returns early (no ask) **and** the delegate returns early (no enforcement) ⇒ that write is governed by nothing. Today, without the widening, the ask still fires, which is the last remaining oversight.

  Measured rather than argued: applying that same widening on top of #1978 fails #1978's own guard test —

  ```
  (fail) tool.assertExternalDirectory > still asks for a memory path named through an aliased data root (deferral stays lexical)
  error: expect(received).toBeDefined()
  Received: undefined
   13 pass
   1 fail
  ```

  It was also inconsistent: it widened 2 of the 4 `<data>/memory` deferral sites, leaving `apply_patch.ts:193` and `:209` lexical.

### Method note for future attribution checks

#1978's ownership check scanned 218 **open PRs** and correctly concluded nobody owned the area — but it could not see this branch, because the branch has no PR. An attribution check restricted to PRs has a blind spot for pushed-but-unopened refs. `git ls-remote origin 'refs/heads/*'` (236 refs here) closes it, and the output must not be piped through `cut` — truncation has already caused one branch name to be misread in this area.

wqymi:

PR #2004, opened 2026-08-01
fix(provider): auto-detect reasoning models from reasoningEffort option

### LINKED ISSUE #2003: fix(provider): auto-detect reasoning models from reasoningEffort option
When users define custom providers with reasoningEffort in model options, they must also manually set reasoning: true. Otherwise, the variants() function returns empty and Ctrl+T toggle doesn't appear.

Auto-detect reasoning models by checking if reasoningEffort is set in options. This way, any model with reasoningEffort automatically gets the reasoning capability flag.

anandlo:
Closes #2003

## What

Auto-detect reasoning models by checking if reasoningEffort is set in options.

## Why

When users define custom providers with reasoningEffort in model options, they must also manually set reasoning: true. Otherwise, the variants() function returns empty and Ctrl+T toggle doesn't appear.

## Changes

Check if reasoningEffort is set in model options when determining the reasoning capability flag.

## Verification

- 421 provider tests pass
- TypeScript typecheck passes
- Change is minimal: 1 file, 1 line changed

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -1203,7 +1203,7 @@ const layer: Layer.Layer<
               providerID: ProviderID.make(providerID),
               capabilities: {
                 temperature: model.temperature ?? existingModel?.capabilities.temperature ?? false,
-                reasoning: model.reasoning ?? existingModel?.capabilities.reasoning ?? false,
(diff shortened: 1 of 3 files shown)


anandlo:
Closing to recreate from clean main branch. This PR incorrectly includes unrelated Hy3 changes.

PR #2007, opened 2026-08-01
feat(provider): surface free-tier opencode models when authenticated

### LINKED ISSUE #2006: feat(provider): surface free-tier opencode models when authenticated
The opencode provider filter hides ALL free models (cost.input === 0) even when the user has a valid API key:

if (!ok || value.cost.input === 0) delete input.models[key]

Users with a subscription can't see free-tier models (mimo-v2.5-free, deepseek-v4-flash-free, etc.) in the model picker, and must manually define them as a custom provider to use them.

Proposal: show free models when the user is authenticated. Unauthenticated users still see nothing.

anandlo:
Closes #2006

## What

Show free-tier opencode models (mimo-v2.5-free, deepseek-v4-flash-free, etc.) when the user is authenticated.

## Why

The current filter hides ALL free models (cost.input === 0) even with a valid API key. Users must manually define free models as a custom provider to use them, and miss out on automatic model metadata (reasoning capability, variants, pricing).

## Changes

Remove the cost-based filter. Unauthenticated users still see no models.

## Verification

- 83 provider tests pass
- TypeScript typecheck passes

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -188,12 +188,10 @@ function custom(dep: CustomDep): Record<string, CustomLoader> {
         Boolean(yield* dep.auth(input.id)) ||
         Boolean((yield* dep.config()).provider?.["opencode"]?.options?.apiKey)
 
-      // Never surface the free/public tier (cost.input === 0). Without a
-      // subscription/key, hide the remaining (paid) models too — they can't be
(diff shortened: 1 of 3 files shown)


anandlo:
Closing to recreate from clean main branch. This PR incorrectly includes unrelated Hy3 and auto-detect changes.

PR #2013, opened 2026-08-02
fix(session): treat GPT empty-output steps as terminal, not invalid

MiMoHardFather:
## Summary
- Fixes the mimorouter GPT bug where the LLM loop terminates right after the model produces output ("empty output" → loop stops).
- Root cause: GPT models behind OpenAI-compatible proxies (mimorouter/LiteLLM) end a tool-loop step with no usable output (no text, no reasoning) because encrypted reasoning items aren't echoed back (`transform.ts` only requests `reasoning.encrypted_content` for `@ai-sdk/openai`). The empty step hit the `invalid "empty output"` fallback in `classify.ts`, which `runLoop` nudged twice and then terminated with a spurious `InvalidOutputError`.
- Fix: classify a GPT-family finished step with no usable output as `final` (degraded on finish `other`), mirroring the existing reasoning-only→final branch, so the loop ends cleanly on the model's own `stop` without a fake error. Non-GPT models are unchanged (still `invalid` → nudge). GPT detection is factored into a shared `isGptFamily()` helper used by both terminal branches.

## Changes
- `packages/opencode/src/session/classify.ts` — GPT empty-output → `final`; shared `isGptFamily()` helper.
- `packages/opencode/test/session/classify.test.ts` — 4 new unit cases (GPT stop/other/namespaced empty → final; non-GPT empty stays invalid).
- `packages/opencode/test/session/invalid-output-continuation.test.ts` — 1 integration case (GPT empty stop step is terminal and not retried).

## Verification
- `bun test test/session/classify.test.ts test/session/invalid-output-continuation.test.ts --timeout 30000` → 42 pass / 0 fail.
- `bun typecheck` clean; pre-push workspace typecheck (12 tasks) green.

Note: integration tests in `invalid-output-continuation.test.ts` need `--timeout 30000` — the empty/think-only cases exceed the 5s default on some machines (env timing, not a logic failure).

### DIFF
--- a/packages/opencode/src/flag/flag.ts
+++ b/packages/opencode/src/flag/flag.ts
@@ -88,6 +88,7 @@ export const Flag = {
   MIMOCODE_DISABLE_MOUSE: truthy("MIMOCODE_DISABLE_MOUSE"),
   MIMOCODE_OUTPUT_LENGTH_CONTINUATION_LIMIT: number("MIMOCODE_OUTPUT_LENGTH_CONTINUATION_LIMIT") ?? 3,
   MIMOCODE_INVALID_OUTPUT_CONTINUATION_LIMIT: number("MIMOCODE_INVALID_OUTPUT_CONTINUATION_LIMIT") ?? 2,
+  MIMOCODE_EMPTY_OUTPUT_RETRY_LIMIT: number("MIMOCODE_EMPTY_OUTPUT_RETRY_LIMIT") ?? 3,
   MIMOCODE_TEXT_TOOL_CALL_RETRY_LIMIT: number("MIMOCODE_TEXT_TOOL_CALL_RETRY_LIMIT") ?? 2,
   // Defaults to false. When enabled, unsigned historical reasoning sent through
   // the Anthropic Messages format receives an empty placeholder signature so it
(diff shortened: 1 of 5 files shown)


MiMoHardFather:
Follow-up (second commit): corrected the fix direction.

The first commit made a GPT empty step terminate cleanly (`final`) — but user testing showed the loop **still stopped after one round**, which is not the goal. The desired behavior is Codex parity: **keep the tool loop going**.

Why Codex works but mimorouter doesn't: the Responses language model sets `hasFunctionCall` and forces unified finish `tool-calls` even when the raw reason is `stop`/null, so classify re-loops. The OpenAI-compatible path (mimorouter/LiteLLM) has no such fallback, so a gpt-5 tool-loop step can land empty (encrypted reasoning isn't echoed back — `include` is openai-only) with finish `stop` and the loop dies.

New approach:
- `classify.ts` returns a new `empty-retry` verdict for GPT-family empty steps (`isGptFamily()`), instead of `final`.
- `runLoop` routes it to a new `autoRetryEmptyOutput` helper that **regenerates** (plain re-request via a minimal synthetic reminder), bounded by `MIMOCODE_EMPTY_OUTPUT_RETRY_LIMIT` (default 3), then terminates with `InvalidOutputError` on exhaustion.
- Under `json_schema` the empty step still routes to structured-retry. Non-GPT empty keeps the nudge-based `invalid` path; GPT reasoning-only still terminates as `final`.

Tests updated: classify unit cases now expect `empty-retry`; the integration test now asserts the empty stop step is **regenerated** (2 requests → real answer, no error). 42 pass, typecheck clean.

PR #2028, opened 2026-08-04
fix(provider): send function tools to OpenAI Responses with explicit strict: false

clyfish:
## Problem

MiMo Code requests to a Codex backend failed **mid-stream** at a 5–17% rate: HTTP 200, some tokens already delivered, then `event: error` (`server_error`) + `response.failed`. To the user the answer just stops half-written. The official `codex` client on the same upstream failed at 0.2–0.9%, and for unrelated reasons (`server_is_overloaded`, context limits).

The official client is immune because **it sends no tools** — its built-ins are provided by the backend. Every client that ships its own tools hits this.

## Root cause

OpenAI's Responses API treats a function tool that **omits** `strict` as strict, and `@ai-sdk/openai` only emits the field when the tool sets it:

```js
// node_modules/@ai-sdk/openai/dist/index.mjs:4446
...tool.strict != null ? { strict: tool.strict } : {},
```

`ai` core has the same guard (`ai/dist/index.mjs:1806`). So we were opting into constrained decoding by accident, never having stated an intent.

Our schemas are deliberately **not** strict-compatible — optional parameters stay out of `required`, not every object carries `additionalProperties: false`, and discriminated unions keep an `anyOf`. Rather than reject the request, the backend auto-patches the schema and then fails to compile the resulting decoding grammar. Observed in the `response.created` echo: all 17 tools came back tagged `strict: true`, `bash.required` grew from 2 entries to 5 (`timeout`/`workdir`/`interactive` all became mandatory), and `task.parameters.properties.operation` gained `additionalProperties: false`.

Because that failure happens at **generation** time, the 200 is already committed, so the error can only surface mid-stream.

What made this expensive to diagnose is the asymmetry: the *same* invalid schema sent with an explicit `strict: true` gets a clean 502 (no partial output, no billing), while omitting the field passes validation and then 500s mid-generation. It reads as random upstream flakiness rather than a deterministic schema problem.

## Fix

State the intent explicitly. **No schema changes** — `anyOf` unions and optional parameters stay exactly as they are.

`ProviderTransform.tools()` is the single choke point for the outbound tool set (`session/llm.ts:733`), so one place covers built-in, MCP, and `StructuredOutput` tools:

```ts
if (EXPLICIT_NON_STRICT_TOOL_SDKS.includes(model.api.npm)) {
  for (const tool of Object.values(tools)) {
    if (tool && tool.strict == null) tool.strict = false
  }
}
```

Scoped to the SDKs that reach an OpenAI Responses endpoint **and** forward `tool.strict`:

- `@ai-sdk/openai`
- `@ai-sdk/azure` — builds `OpenAIResponsesLanguageModel` from `@ai-sdk/openai/internal`

Deliberately **not** applied elsewhere: the vendored Copilot Responses SDK already always emits `strict` (`sdk/copilot/responses/openai-responses-prepare-tools.ts:50`), and `@ai-sdk/anthropic` warns for *any* non-null `strict` (`dist/index.mjs:1505`), so a blanket default would spam warnings on every Anthropic request. An explicit per-tool `strict` is preserved so a tool that is made strict-compatible can still opt in.

## Second commit: the same bug class on the structured-output path

While verifying the scope I found the `response_format` sibling. `generateObject`/`streamObject` default `strictJsonSchema` to **true** (`@ai-sdk/openai` chat + responses, and `@ai-sdk/openai-compatible`), so `strict: true` goes out *explicitly*.

`SessionGoal.Verdict` marks `impossible` optional, so `required` ships 2 of its 3 properties. Captured from the wire:

```
strict: true
properties: ["ok", "impossible", "reason"]
required:   ["ok",               "reason"]   ← strict mode rejects this
```

`goal.ts` judges with the **session's** model, so the stop-condition judge (`/goal`) was broken on every OpenAI-backed model. This one fails cleanly at validation rather than mid-stream, so it is a separate, visible bug — but the root cause is identical.

Fixed the same way rather than by rewriting the schema: `impossible` is optional *by design* (JUDGE_SYSTEM tells the judge to return `{"ok": false}` **without** `impossible` when in doubt), so forcing it into `required` would change what the judge is asked to produce. `.default(false)` does not help — it still lands outside `required`.

The agent-config schema in `agent/agent.ts` is deliberately **not** opted out: it is already strict-compatible (all fields required, `additionalProperties: false`), so it keeps constrained decoding. A test pins that so adding an optional field there fails loudly instead of silently reintroducing this bug.

## Testing

`test/provider/strict-schema-wire.test.ts` (13 tests) follows the existing `empty-content-wire.test.ts` pattern: inject `fetch` and assert on the **real outbound HTTP body**, driven through the full chain (`ToolSet` → `ai` core → provider → wire). Asserting on `ProviderTransform.tools()`'s return value alone would be two layers short — `ai`'s `prepareToolsAndToolChoice` has the same `tool.strict != null` guard, so a field that failed to survive it would never reach the provider.

Includes **CONTROL** cases proving the defect mechanism (and therefore that these tests would catch a regression): untransformed tools ship with no `strict` field at all, and `Verdict` ships `strict: true` with an incomplete `required`.

Also covered:
- schemas byte-identical before/after — only `strict` is added
- MCP tools (`dynamicTool`, `type: "dynamic"`, which `ai` funnels through the same `case "function"` branch) — server-supplied schemas are the least likely to be strict-compatible
- Anthropic keeps `strict` unset **and** emits no `feature: "strict"` warning
- an explicit `strict: true` survives

Results:
- `bun test test/provider/` → **471 pass, 0 fail**
- `bun test test/session/` → **898 pass, 25 skip, 0 fail**
- `bun typecheck` → clean

## Notes for reviewers

Both transforms **mutate in place**. That is safe because the record and every tool object in it are rebuilt per request: `resolveTools` allocates a fresh record and calls `tool()` per entry, and MCP entries come from `convertMcpTool`, which returns a new `dynamicTool()` on every `MCP.tools()` call. Nothing outlives the request, so a mid-turn model switch cannot carry `strict` over to a provider that would reject or warn on it. This is documented at the function, since it is the correctness precondition for the approach.

Azure's `useCompletionUrls` branch sends the same tools to Chat Completions, where the field lands as `function.strict: false` — a documented boolean whose default is already false, so that path is unaffected.

Not a regression: `transform.ts` and the pinned `@ai-sdk/openai@3.0.53` both date to the initial open-source commit. It is the interaction of three individually-reasonable behaviours (the SDK not inventing an intent, OpenAI defaulting to strict, our schemas being intentionally non-strict) that only becomes visible when tools are sent to a Codex backend.

**Known remaining gap:** the vendored Copilot Responses SDK defaults `strictJsonSchema` to `false` (`openai-responses-language-model.ts:212`) while the npm packages default it to `true`, so structured output behaves differently across providers. Left alone here to keep this PR scoped; worth unifying separately.

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -1039,17 +1039,37 @@ export function message(msgs: ModelMessage[], model: Provider.Model, options: Re
 //
 // An explicit per-tool `strict` is preserved, so a tool that has been made
 // strict-compatible can still opt in.
+//
+// Azure's `useCompletionUrls` branch sends the same tools to Chat Completions
+// instead, where the field lands as `function.strict: false` — a documented
+// boolean whose default is already false, so that path is unaffected.
 const EXPLICIT_NON_STRICT_TOOL_SDKS = ["@ai-sdk/openai", "@ai-sdk/azure"]
 
-// Place a cache breakpoint on the tool definitions. The cache hierarchy is
-// `tools` → `system` → `messages`, so marking the LAST tool caches the entire
-// tool-schema block (often several KB) as a stable prefix that sits in front of
-// the system + message caches. Tools are passed to the SDK separately from
-// `message()` and never go through its providerID→SDK-key remap, so we resolve
-// the SDK-keyed marker via `cacheMarkerFor`. Tool registration order is stable
-// (insertion order of the tools record), so "last tool" is deterministic.
+// The single choke point for the outbound tool set (session/llm.ts passes the
+// result straight to `streamText`). Two responsibilities:
+//
+// 1. Pin `strict: false` for EXPLICIT_NON_STRICT_TOOL_SDKS — see above for why
+//    omitting the field breaks the Codex backend mid-stream.
+// 2. Place a cache breakpoint on the tool definitions. The cache hierarchy is
+//    `tools` → `system` → `messages`, so marking the LAST tool caches the entire
+//    tool-schema block (often several KB) as a stable prefix that sits in front
+//    of the system + message caches. Tools are passed to the SDK separately from
+//    `message()` and never go through its providerID→SDK-key remap, so we
+//    resolve the SDK-keyed marker via `cacheMarkerFor`. Tool registration order
+//    is stable (insertion order of the tools record), so "last tool" is
+//    deterministic.
+//
+// Both mutate in place. That is safe because the record and every tool object in
+// it are rebuilt per request: `resolveTools` allocates a fresh record and calls
+// `tool()` per entry, and MCP entries come from `convertMcpTool`, which returns
+// a new `dynamicTool()` on every `MCP.tools()` call. Nothing here outlives the
+// request, so a model switch between steps cannot carry `strict` over to a
+// provider that would reject or warn on it.
 export function tools<T extends Record<string, any>>(tools: T, model: Provider.Model): T {
   if (EXPLICIT_NON_STRICT_TOOL_SDKS.includes(model.api.npm)) {
+    // Guarded because this walks every entry; the single `last` lookup below can
+    // assume a well-formed record, but a loop over N values is cheaper to make
+    // safe than to debug as a crash in the request path.
     for (const tool of Object.values(tools)) {
       if (tool && tool.strict == null) tool.strict = false
     }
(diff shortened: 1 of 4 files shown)


clyfish:
Both review findings addressed in 807ff9e2.

**Finding 2 — unconditional `providerOptions` in goal.ts: fixed.** `structuredOutputOptions` now returns `undefined` instead of `{}`, so the non-OpenAI path attaches no provider-options bag at all. You were right that it was harmless, but it was an unintended new field on providers this change has no business touching.

**Finding 1 — xai: investigated, and I'm leaving it out, with the reasoning now in the code.**

The structural half of your point is correct and I've fixed that: the list is keyed by npm while the Responses-vs-Chat decision is made per *provider* in `getModel`, so the two can drift silently. The comment now enumerates every `sdk.responses()` call site with a verdict:

| call site | npm | verdict |
|---|---|---|
| `:323` openai | `@ai-sdk/openai` | listed |
| `:359` azure | `@ai-sdk/azure` | listed |
| `:379` azure-cognitive-services | `@ai-sdk/azure` | covered — catalog pins the provider npm, confirmed |
| `:340` github-copilot | → vendored `./sdk/copilot` | immune, already always emits `strict` |
| `:331` xai | `@ai-sdk/xai` | deliberately excluded |

On xai specifically — it does forward `tool.strict` behind the same omit-when-null guard, so it genuinely looks like it belongs. But its prepare-tools runs every tool schema through `removeAdditionalPropertiesFalse`:

```js
// @ai-sdk/xai/dist/index.mjs:319 — called at :2035, inside prepareResponsesTools
if (key === "additionalProperties" && propertyValue === false) continue
```

Strict mode *requires* `additionalProperties: false` on every object. So if xAI's Responses API were strict-by-default, the xai SDK would reject every tool call it makes, by construction. It can't be strict by default — and forcing `strict: false` there would assert a constraint xAI hasn't been shown to honour.

So the exclusion isn't "no evidence either way" (which was my weaker original reason) — it's positive evidence against. Pinned by a test so it reads as a decision rather than an oversight; if xai ever does need this, that test is where the conversation starts.

Re-verified after the change: `test/provider/` **472 pass / 0 fail**, `test/session/` **898 pass / 25 skip / 0 fail**, `bun typecheck` clean.

clyfish:
Fixed in e77356b7 — and I took the substantive fix rather than the rename, because you'd found the one claim in this PR that was asserted by citation instead of by test.

The `additionalProperties` stripping is the *load-bearing premise* of the xai exclusion. If it's wrong, xai belongs in the list and there's a live mid-stream bug. So it shouldn't rest on a source citation in a comment when everything else here is proven at the wire.

Split into two tests:

1. `xai tools are left untouched, so nothing reaches the wire` — the exclusion itself, which is what the old assertions actually covered.
2. `xai strips additionalProperties: false, which strict mode requires` — sends a schema that **does** carry the field and asserts xai drops it, with the OpenAI SDK as a contrast shipping the same schema intact. That second half matters: it shows the stripping is xai-specific rather than something `ai` core does upstream, which the single-provider version couldn't distinguish.

You were exactly right about why the old test couldn't have caught it — `toolset()` has no `additionalProperties: false` to strip, so the assertion was structurally incapable of observing the behaviour it was named for.

The useful property now: if xai ever stops stripping the field, test 2 fails and points a maintainer straight at the exclusion decision, instead of the reasoning quietly going stale in a comment.

`test/provider/` **473 pass / 0 fail** (15 in this file), `bun typecheck` clean, prettier clean.

Standing caveat unchanged: the xai conclusion is still inference from SDK behaviour, not a request against xAI's API. A single tool-bearing call with explicit `strict: true` from anyone holding an xai key would settle it empirically.

PR #2101, opened 2026-08-12
fix(auth): stop handing user credentials to child processes

yuyang27:
`MIMOCODE_AUTH_CONTENT` (the whole auth.json: provider keys, OAuth refresh tokens) and `MIMOCODE_CONFIG_CONTENT` (can embed MCP `environment` values and request `headers`) travelled in the environment, and every child the engine spawns inherited them wholesale. An agent does not even need to know where auth.json lives — `echo $MIMOCODE_AUTH_CONTENT` from the bash tool is enough.

Affected spawn paths:

- `tool/bash.ts` and the shell part in `session/prompt.ts` — agent-authored commands;
- local MCP servers (`mcp/index.ts`) — third-party binaries the user installed;
- LSP servers, the PTY, ripgrep, formatters, the worktree start command, git/snapshot helpers;
- `effect/cross-spawn-spawner.ts`, the funnel every `extendEnv: true` caller goes through;
- `util/process.ts`, which inherited implicitly whenever no env was passed.

Two changes:

1. `withoutCredentials()` scrubs both vars from the *inherited* half of any env handed to a child. A caller that passes credentials explicitly is left alone, so control-plane workspaces (`control-plane/workspace.ts` builds its own env) and the TUI worker keep working.

2. `Auth.inject()` gives an embedding host an in-process channel so credentials need not be in the environment at all. Scrubbing the child's own env is not sufficient by itself: a child running as the same user can read `/proc/<parent-pid>/environ`. The env var stays as a fallback.

Behavior change worth noting: `opencode` configured as its own MCP server no longer inherits the host's credentials or config content; that nested engine falls back to reading auth.json and the config file from disk, which is what a plain CLI invocation does.

Tests (`test/util/credential-env.test.ts`): the helper, plus two source scans — one that no spawn site hands the inherited env over unscrubbed, one that `sanitizedProcessEnv` (a full copy of the environment despite the name) is only used where the child is an engine process. The scan pattern covers `...globalThis.process.env`, not just `...process.env`: the spawner funnel is written that way, and a narrower pattern reports a clean tree while the widest leak path stays open. Flip-verified: reverting the funnel or the ripgrep call site turns the scans red at the exact line. `test/auth/auth.test.ts` covers inject precedence and fallback.

### Type of change

Bug fix / New feature / Refactor / Documentation — keep the ones that apply and delete the rest.

### What does this PR do?

Briefly describe the problem, what changed, and why this approach works.

What did you test, and how can a reviewer reproduce the result?

### Screenshots / recordings

_If this is a UI change, please include a screenshot or recording._

### Checklist

- [ ] I have tested my changes locally
- [ ] I have not included unrelated changes in this PR

### DIFF
--- a/packages/opencode/src/auth/index.ts
+++ b/packages/opencode/src/auth/index.ts
@@ -8,6 +8,23 @@ export const OAUTH_DUMMY_KEY = "mimocode-oauth-dummy-key"
 
 const file = path.join(Global.Path.data, "auth.json")
 
+/**
+ * Credentials supplied by an embedding host (the desktop app runs the engine in-process).
+ *
+ * Kept in memory on purpose: the previous channel was `MIMOCODE_AUTH_CONTENT`, and anything in
+ * the environment is readable by every child the engine spawns — the bash tool alone turns
+ * `echo $MIMOCODE_AUTH_CONTENT` into a credential dump, and a sibling can still read
+ * `/proc/<pid>/environ` even after the child's own env is scrubbed. A module-level value has no
+ * such surface. The env var is still honored so workspace children keep working (they receive it
+ * explicitly), but hosts should prefer `inject`.
+ */
+let injected: string | undefined
+
+/** Set (or clear, with `undefined`) the in-process credentials. Read on every `all()`. */
+export function inject(content: string | undefined) {
+  injected = content
+}
+
 const fail = (message: string) => (cause: unknown) => new AuthError({ message, cause })
 
 export class Oauth extends Schema.Class<Oauth>("OAuth")({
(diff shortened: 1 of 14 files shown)


MiMoHardFather (changes requested):
This security fix still has merge-blocking gaps:

1. Project config can re-inject the scrubbed variables. MCP/LSP/formatter environments are resolved from config (including `{env:MIMOCODE_AUTH_CONTENT}`) and merged *after* `withoutCredentials(process.env)`, so an untrusted repository can restore both credential variables in an attacker-controlled child. Please scrub the final merged environment, with a separate narrow opt-in for trusted engine/control-plane children.

2. Direct `child_process`/`open` call sites that omit `env` still inherit both variables (for example the interactive sqlite3 process in `src/cli/cmd/db.ts`). The new source scan explicitly does not catch omitted-env spawns. Please route these through the sanitized wrapper or enforce a structural rule preventing unaudited direct spawns.

3. `Auth.inject()` remains authoritative after `Auth.set()`/`remove()`, so login, key rotation, or logout writes succeed but subsequent reads keep returning the old injected snapshot. The injected state and mutation paths need consistent semantics.

Also, `MIMOCODE_CONFIG_CONTENT` still has no in-process injection channel, despite containing secrets and being readable from `/proc/<parent>/environ` on Linux.

Please add regression tests for config-based re-injection, omitted-env spawn coverage, and auth mutation while injection is active.

MiMoHardFather (approved):
Approved for merge per maintainer direction; known review findings are accepted for follow-up.

PR #2104, opened 2026-08-12
fix(auth): make writes visible to reads while credentials are injected

yuyang27:
`Auth.all()` returns the injected snapshot (or the `MIMOCODE_AUTH_CONTENT` fallback) and never consults the file when one is present, but `set`/`remove` only wrote the file. So while either channel is active, a write landed on disk and every later read in that process kept returning the pre-write state: an OAuth token refresh, a key rotation or a logout looked like it succeeded and changed nothing until the process restarted.

The env channel had this defect before `inject` existed — a refresh inside a workspace child was invisible to that same child.

A mutation through this service is authoritative, so the snapshot now moves with the write. It is updated only *after* the write lands: moving it first would leave the in-memory credentials ahead of the file on a failed write, and the next process would silently run on the old ones. An embedding host that re-injects from the file on its own schedule converges on the same content.

Tests: `set`/`remove` visible to later reads with `inject` active, the same for the env channel, and a failed write leaving the snapshot untouched (flip-verified — moving the update before the write turns that one red).

### Type of change

Bug fix / New feature / Refactor / Documentation — keep the ones that apply and delete the rest.

### What does this PR do?

Briefly describe the problem, what changed, and why this approach works.

What did you test, and how can a reviewer reproduce the result?

### Screenshots / recordings

_If this is a UI change, please include a screenshot or recording._

### Checklist

- [ ] I have tested my changes locally
- [ ] I have not included unrelated changes in this PR

### DIFF
--- a/packages/opencode/src/auth/index.ts
+++ b/packages/opencode/src/auth/index.ts
@@ -25,6 +25,9 @@ export function inject(content: string | undefined) {
   injected = content
 }
 
+/** Whether reads currently come from a snapshot rather than the file. */
+const snapshotActive = () => injected !== undefined || process.env.MIMOCODE_AUTH_CONTENT !== undefined
+
 const fail = (message: string) => (cause: unknown) => new AuthError({ message, cause })
 
 export class Oauth extends Schema.Class<Oauth>("OAuth")({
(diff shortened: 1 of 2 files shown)


yuyang27:
@MiMoHardFather  麻烦review下

PR #2125, opened 2026-08-14
fix(session): order messages by (time_created, id) so an id wrap can't wedge a session

JinyuXiang-Mimo:
## Summary

Sessions with history from before **2026-08-14 11:19:55 UTC** silently stop responding: a new prompt is stored, no provider call is ever made, and nothing surfaces to the user.

`Identifier.create` packs `Date.now() * 0x1000 + counter` into 6 bytes, so the sortable id prefix wraps every `2^36 ms` (~2.18 years). The last boundary was `1786706395136`, which makes every post-wrap id (`msg_000…`) sort **before** the entire pre-wrap history (`msg_fff…`). `Session.lastMainMessageID` then returns a days-old message as the tail, the run loop sees no new work, and exits at step 0.

Observed on a real session — three user messages stored on Aug 14, zero assistant replies, last processed turn Aug 06:

```
msg_fd708d21e001JXYNUE1Jba3VEw  assistant  2026-08-06 20:25  ← id sorts LAST
msg_0006f768700114fd6bDaDwzOWs  user       2026-08-14 21:21  ← id sorts FIRST
```

## Approach

Fix the **ordering**, not the encoding. `time_created` is an independent integer column, unaffected by the 48-bit packing, so it becomes the primary sort key; the id only breaks ties within the same millisecond — exactly the ambiguity the counter exists to resolve.

This needs **no migration** and unwedges existing sessions immediately. Widening the encoded value is still worth doing for the ~Oct 2028 boundary, but it would **not** restore order on its own: pre-wrap `ff…` ids keep sorting above freshly minted ones. A follow-up can widen the encoding; this PR makes correctness independent of it.

Adds `MessageV2.compare(a, b)` as the single chronological comparator and routes the affected call sites through it.

## Sites corrected

Each was checked for intent before changing — several encode "later in the stored sequence" rather than "later in time".

| Site | Consequence if inverted |
|---|---|
| `session.ts` `lastMainMessageID` | **The wedge.** `desc(id)` → `desc(time_created), desc(id)` |
| `session.ts` `fork` | Truncation cutoff — forks an empty session or copies all of it |
| `revert.ts` `cleanup` / `revert` | These **delete** what they classify as "after" the revert point — inverted, they take the history instead of the tail |
| `prompt.ts` ×3 | Auto-continue-on-length, the `step > 1` user-prompt reminder scan, the fork-agent watermark filter |
| `history/backfill.ts` | `gt(id, cursor)` walk → composite `(time_created, id)` cursor; otherwise post-wrap parts are never indexed into FTS |

### One site deliberately left alone

`usageRecovered` in `prompt.ts` is **not** converted to a compare. `msgs` comes from `filterCompacted`, which walks newest-first and stops at the first checkpoint/compaction marker — so the slice holds at most one marker and it is always `msgs[0]`. "A marker exists in this slice" already means "this turn was rebuilt".

Both orderings that look correct there are wrong: a position compare never fires (nothing follows the marker), and a time compare never fires either (the marker carries the synthetic time `boundary.time.created + 1`, so it sorts *earlier* than `lastFinished`). The old `id >` worked only by accident — a freshly minted id sorts highest — and stops working across a wrap.

Getting this wrong un-guards the overflow path and rebuilds the same turn twice; that is how it surfaced in `auto-overflow-writer-first.test.ts` during development. The reasoning is now recorded at the call site so the next reader doesn't "fix" it.

## Testing

- `bun typecheck` — clean (all 12 packages)
- New `test/session/message-id-wrap-order.test.ts` — 7 cases, using real ids from an affected session. One case asserts the encoder **still wraps**, so a later widening fails loudly here rather than leaving this fix looking unmotivated.
- `test/session/` — 917 pass. 4 failures are unrelated: 2 fail identically on a clean baseline (`checkpoint-child-session` T10, `prune.test.ts` recovery gate), and 2 are full-run timeout flakes (`a beforeEach/afterEach hook timed out`) that pass 66/66 in isolation on both baseline and this branch. Before this change the same suite had 7 failures.
- `revert-compact`, `fork-prefix-invariant`, `checkpoint-fork-mode`, `main-runloop-history-invariant`, and all of `test/history/` — 52 pass, 0 fail.

## Note

The next wrap is **~Oct 2028**. Two follow-ups worth considering: widen the encoded value, and assert the invariant (`new user message id < session max id`) so "silently ignored prompt" stops being a possible failure mode at all — that, more than the ordering bug, is what made this expensive to diagnose.

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/context/sync.tsx
+++ b/packages/opencode/src/cli/cmd/tui/context/sync.tsx
@@ -158,6 +158,85 @@ export function bucketMessages<M extends { agentID?: string | null }>(
 }
 
 /**
+ * Client-side mirror of MessageV2.compare — chronological order of two messages.
+ *
+ * Message ids are NOT a clock. The id encoder packs `Date.now() * 0x1000 + counter`
+ * into 6 bytes, so the sortable prefix wraps every 2^36 ms (~2.18 years); the last
+ * boundary was 2026-08-14T11:19:55Z. Across a wrap a NEWER message gets a SMALLER
+ * id, so `a.id < b.id` inverts for any session whose history straddles it.
+ *
+ * This matters more here than it looks: the store keeps each bucket sorted and uses
+ * Binary.search over that order both to locate an existing message and to pick the
+ * splice index for a new one. With ids as the key, a fresh message lands at index 0
+ * — which is what put new messages at the top of the transcript — and the >100
+ * trim then treats `list[0]` as "oldest" and deletes it along with its parts.
+ *
+ * `time.created` is an independent field, unaffected by the packing, so it is the
+ * primary key; the id only breaks ties inside the same millisecond.
+ */
+export function compareMessages(
+  a: { id: string; time: { created: number } },
+  b: { id: string; time: { created: number } },
+) {
+  return a.time.created !== b.time.created ? a.time.created - b.time.created : a.id < b.id ? -1 : a.id > b.id ? 1 : 0
+}
+
+/**
+ * Compare a message against a MARKER id (revert point, pending watermark) held by
+ * the session rather than by a message object.
+ *
+ * Returns `< 0` if `msg` precedes the marker, `> 0` if it follows, `0` if it IS the
+ * marker. When the marker is not present in `list` there is no anchor to compare
+ * against and the caller gets `undefined` — callers decide what that means rather
+ * than silently treating the marker as position 0 or infinity.
+ *
+ * Resolving the marker through the list is what makes this wrap-safe: the marker id
+ * alone carries no time, and comparing id strings inverts across a wrap boundary
+ * (see compareMessages).
+ */
+export function compareToMarker<M extends { id: string; time: { created: number } }>(
+  list: M[],
+  msg: M,
+  markerID: string,
+): number | undefined {
+  const marker = list.find((m) => m.id === markerID)
+  if (!marker) return undefined
+  return compareMessages(msg, marker)
+}
+
+/**
+ * Locate `id` in a chronologically-sorted message list, or the index it should be
+ * inserted at. Same contract as Binary.search, but keyed on (time.created, id) —
+ * Binary.search compares id strings only and cannot express this order.
+ *
+ * `probe` is the incoming message when inserting. Lookups that only have an id
+ * (message.removed) pass none and fall back to a linear scan: without a time there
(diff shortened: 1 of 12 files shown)


JinyuXiang-Mimo:
Good catch — confirmed and fixed in 792cdfba. `classify.ts` was a real miss, and it mattered more than the ones already covered: `classifyAssistantStep` is called from all three runLoop classification sites, i.e. the same loop this PR is fixing.

Both guards were converted:

- **#4** (`classify.ts:90`) — across a wrap a newer assistant carries a smaller id, so a live reply is judged stale and the loop re-enters instead of terminating.
- **#3a** (`classify.ts:74`) — same inversion skips detection and falls through to the unconditional tool-calls continue at #3, losing the text-tool-call retry.

## The fixtures were the interesting half

Converting the guards alone would have looked green and proved nothing. Every `userInfo`/`assistantInfo` hardcoded `time: { created: 0 }` and expressed ordering purely through ids (`"m-1"` vs `"m-2"`), so under `MessageV2.compare` all 31 cases would tie on time and silently fall back to the id tie-break — still passing, no longer exercising the ordering they claim to test. `created` now mirrors the id's numeric suffix so fixtures order under `compare` the way they read on the page.

Added 4 cases using real straddling ids, covering **both** directions so the fix can't be a one-way overcorrection:

| Case | Bare id | compare |
|---|---|---|
| id order is inverted for this pair (premise) | — | asserts `postWrap < preWrap` |
| fresh post-wrap assistant not judged stale | `continue` ❌ | `final` ✅ |
| genuinely stale pre-wrap assistant still stale | ❌ | `continue` ✅ |
| text-tool-call in fresh post-wrap turn detected | falls to #3 ❌ | `text-tool-call` ✅ |

Verified by reverting `classify.ts` to the bare-id version: exactly those 3 behavioural cases fail, 32 pass. With the fix: 35 pass.

## Rescan

Re-swept the package for id-vs-id comparisons and drizzle-level `orderBy`/`lt`/`gt` on `MessageTable.id` / `PartTable.id`. Nothing else outstanding.

Two `orderBy(PartTable.id)` sites (`message-v2.ts:640`, `:1049`) are left deliberately: both order parts **within a single message**, which are minted in one wrap epoch, so their relative order holds. Noting it explicitly rather than silently skipping — happy to convert them for uniformity if you'd prefer, but it's churn without a defect behind it.

`bun typecheck` clean; `classify` + `classify-integration` + `message-id-wrap-order` = 51 pass, 0 fail.

PR #2138, opened 2026-08-16
feat(agent): make one level of subagent delegation the default

yanyihan-xiaomi:
## Summary

A `general` subagent could spawn further subagents with **no depth limit**: it carries `defaults`' `"*": "allow"`, declares no `toolAllowlist`, and nothing on the spawn path checks ancestry (`Actor.spawn` records only `parentActorID`). Its prompt actively invited it — "You may delegate genuinely independent, bounded subtasks…" — so agent trees could fan out unbounded with no single cap governing cost or concurrency.

This makes **one level of delegation the default**: a subagent no longer sees the `actor` tool at all.

- `src/agent/agent.ts` — a pass at the end of `Agent.state` appends `actor: "deny"` to every `mode === "subagent"` agent. Because the pattern is `"*"`, `Permission.disabled` drops `actor` from that agent's LLM-visible schema in `llm.ts` `resolveTools`, which also makes its `ctx.ask({ permission: "actor" })` unreachable.
- The rule must be **appended last**: `evaluate`/`disabled` use `findLast`, and `defaults`' wildcard `"*": "allow"` matches the `actor` tool name, so a rule merged earlier loses to it. The opt-back-in is therefore the skip guard — an explicit `actor` rule (`agent.<name>.permission.actor: "allow"` or a global `permission.actor`) suppresses the injection, mirroring the neighbouring `external_directory` pass's "unless explicitly configured" contract. A blanket `"*": "allow"` deliberately does not re-enable nesting.
- `prompt/general.txt` — the delegation bullet is gone. The prompt never names the capability: explaining that delegation is unavailable would plant the concept and send the model looking for the tool.
- `src/session/prompt.ts` — the per-turn recall reminder named `actor({operation:"status"})` in every session with memory or tasks, ungated by agent mode, i.e. it pointed subagents at a tool they no longer hold. `recallHintLines` now takes a `hasActor` flag computed from the acting agent's own visibility.

Programmatic spawn paths are untouched by design, since none of them reads the subagent's tool schema: `bypassAgentCheck` (`@agent` mentions), the workflow runtime's `spawnRef` spawns, `checkpoint-writer` forks, orchestrator peers via the `session` tool. `exec`/tool-script can't re-expose it either — `actor` is in `TOOL_SCRIPT_EXCLUDED`. Permission rules gate the LLM's tool schema, not the spawn API.

`checkpoint-writer` keeps `actor` visible on the fork path: its visibility ruleset is `merge(writer.permission, parentPermission)` with `parentPermission` last, so the parent primary's `"*": "allow"` out-ranks the injected deny and prefix-cache parity holds. A test pins this.

Design rationale, the accepted breadth of the opt-back-in, and the journey log live in `docs/compose/spec/subagent-no-nesting.md` rather than in inline comments. No change to `packages/web/src/content/docs/` — that docs site is upstream content this fork does not maintain.

## Test plan

- `bun typecheck` (packages/opencode) — PASS
- `bun lint` (oxlint) — 0 errors
- `bun test test/agent/ test/tool/actor.test.ts test/tool/actor-recover.test.ts test/actor/` — 233 pass / 9 fail; the identical 9 failures reproduce on clean `main` (PRE-EXISTING: a local global config sets `edit: "ask"`, which un-denies edit-family tools for read-only agents)
- `bun test test/session/prompt-effect.test.ts test/session/recall-reminder.test.ts test/tool/actor.test.ts` — 74 pass, 0 fail
- 5 new tests: subagent-wide `actor` deny with primaries as the negative control, config opt-back-in, `checkpoint-writer` fork parity, absence of the old prompt invitation, recall-hint omission. The wider "prompt never names the tool" property is design intent, not an assertion — it is prose-shaped and would fail on any legitimate rewording.
- Live `bun dev debug agent` — before: `general actor=true`; after: `build actor=true`, `general actor=false`, `explore actor=false`

## Review note

`7b5f766d` (recall-hint gate) landed after the reviewed head `3c2527bf`, as did the comment/prompt cleanup and this revert. All are typechecked, linted and unit-tested, but were not put through a second review pass.

### DIFF
--- a/docs/compose/spec/subagent-no-nesting.md
+++ b/docs/compose/spec/subagent-no-nesting.md
@@ -1,0 +1,66 @@
+---
+feature: subagent-no-nesting
+status: designed
+updated: 2026-06-18
+branch: subagent-no-nesting
+commits: <base-sha>..<head-sha>
+---
+
+# Subagent nesting off by default
+
+## Report
+
+## [S1] Problem
+
+A `general` subagent can spawn further subagents, with no depth limit.
+
+- `agent.ts` gives `general` `Permission.merge(defaults, user)`, and `defaults` carries `"*": "allow"`; it declares no `toolAllowlist`.
+- `Permission.disabled` (permission/index.ts:669) only drops a tool when a matching rule has `pattern === "*"` and `action === "deny"`, so nothing removes `actor` from `general`'s schema.
+- `tool/actor.ts` passes `tools: "INHERIT"` for an agent without a `toolAllowlist`, and `prompt.ts:1276` turns `INHERIT` into "no runtime whitelist".
+- `Actor.spawn` records only `parentActorID`; there is no depth field and no ancestry check anywhere on the spawn path.
+- `prompt/general.txt` actively invites it: "You may delegate genuinely independent, bounded subtasks when that improves throughput".
+
+Verified empirically before the change: `bun dev debug agent general` reports `"actor": true`, `bun dev debug agent explore` reports `"actor": false` (explore only because of its own blanket `"*": "deny"`, not by design).
+
+Consequences: unbounded fan-out of agent trees, cost and concurrency that no single cap governs (`workflow/runtime.ts:34`'s lifecycle cap only bounds `workflow()`'s own `spawnRef` spawns), and a parent that cannot account for work done two levels down.
+
+## [S2] Design
+
+Make one level of delegation the default: a subagent does not see the `actor` tool at all.
+
+**Mechanism.** In `agent.ts`, after the config merge loop and alongside the existing `whitelistedDirs` normalization pass, inject `actor: "deny"` for every agent with `mode === "subagent"`. The rule is merged **before** the agent's accumulated ruleset, so `agent.<name>.permission.actor: "allow"` in user config still wins (`evaluate` uses `findLast`).
+
+```ts
+for (const name in agents) {
+  if (agents[name].mode !== "subagent") continue
+  agents[name].permission = Permission.merge(
+    Permission.fromConfig({ actor: "deny" }),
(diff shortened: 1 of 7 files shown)


yanyihan-xiaomi:
Superseded by #2139, which landed the same invariant in `a9967f52`. Closing this in its favour — its registry-level mask plus the runtime refusal in `ActorTool` is the stronger enforcement, and the `SYSTEM_SPAWNED_AGENT_TYPES` exemption is more explicit than relying on `parentPermission` ordering for `checkpoint-writer`.

Two deltas from this branch are NOT covered by #2139, recorded here for whoever wants to pick them up:

1. **The recall reminder still advertises `actor` to subagents.** `recallHintLines` (`src/session/prompt.ts:132`) and its injection site (~`:3349`) are ungated by agent mode, so any session with memory or tasks appends `- actor({ operation: "status", actor_id: "<id>" })` to the last user message — including a subagent turn, where the tool is now masked out of the schema. That costs a wasted call plus invalid-tool recovery. `skill-search-reminder.ts` gates the analogous hint; this one should too. The fix here was a `hasActor` flag computed from the acting agent's own visibility (commit `7b5f766d`).
2. **`permission.actor` is now a silent no-op for subagents.** The mode-based mask runs regardless of the ruleset, so `agent.<name>.permission.actor: "allow"` no longer has any effect for a `mode: "subagent"` agent, while the config key still validates. Either intended (no escape hatch by design) or worth a schema/doc note.

Design notes, the merge-order trap (`Permission.disabled` uses `findLast`, so an `actor: "deny"` merged before `defaults`' `"*": "allow"` silently does nothing) and the journey log are in `docs/compose/spec/subagent-no-nesting.md` on this branch if any of it is useful later.

PR #2151, opened 2026-08-18
feat(permission): decouple skip-all from permission ask timeout

yanyihan-xiaomi:
## Summary

- Split the monolithic `skipAll` boolean into two orthogonal runtime flags: `skipAll` (auto-allow normal permission asks) and `permissionAskTimeoutMs` (timeout for all human-confirmation asks)
- Added `/permission-timeout` TUI slash command with a modal selector (Never / 30s / 1min / 2min / 5min / 10min)
- Added GET/POST `/instance/permission/ask-timeout` server API endpoints
- Full i18n coverage across 7 languages (en/zh/zht/ru/ja/fr/es)
- Fixed stale skip-permissions toast messages that referenced a hardcoded 60s timeout

## Motivation

The old design coupled skip-all and forced-ask timeout into a single `skipAll` toggle: turning on skip-all also implicitly enabled a 60s auto-reject timeout for `bash_delete` and other forced-ask permissions. This caused a specific failure mode — when a user enabled skip-all expecting the model to freely perform all actions, destructive operations like `rm` would silently auto-reject after 60s of no human response, making the model believe it was unable to delete files. This contradicts the intent of skip-all (trust the model to proceed autonomously).

The two concerns are design-level orthogonal:
- **skip-all** answers "should normal permission asks be auto-allowed?"
- **timeout** answers "how long should the system wait for human confirmation before auto-rejecting?"

They should be independently configurable. Previously, skip-all implicitly forced a 60s timeout; now each is opt-in separately.

## Behavioral change (intentional)

| Before | After |
|--------|-------|
| `skipAll=true` → forced-ask auto-rejects after 60s | `skipAll=true` → forced-ask waits indefinitely (unless timeout is also set) |
| No way to timeout without skip-all | Timeout works independently of skip-all |

The old default of 60s under skip-all was itself a design error — it silently changed skip-all's meaning from "trust the model" to "trust the model, except destructive ops get rejected in 60s." The new default (`null` = no timeout) is the correct conservative baseline: no implicit behavior changes.

## Test plan

All 147 permission tests pass, including 5 new tests covering:
- Env var backward compat (`MIMOCODE_SKIP_ALL_FORCED_ASK_TIMEOUT_MS`)
- Independent timeout without skip-all
- Timeout applies to normal asks (not just forced-ask)
- `null` timeout disables auto-reject
- skip-all + timeout combination


JinyuXiang-Mimo:
666,LGTM

PR #2174, opened 2026-08-20
fix(provider): treat missing/null finish_reason as stop instead of other

### LINKED ISSUE #2173: Bug: models that omit finish_reason are misclassified as finish="other" (degraded/think-only)
## Summary

Models served by the OpenCode Go gateway (`opencode.ai/zen/go/v1`) do **not** emit a `finish_reason` in streaming responses (and return `null` in non-streaming). When `finish_reason` is missing, opencode falls through to `default: return "other"`, which causes `classify.ts` to flag every such response as **degraded** / **think-only** — even though the response was delivered completely and normally.

## Affected model

- `muse-spark-1.2-contributor` via `opencode.ai/zen/go/v1`

## Reproduction / Evidence

Live streaming results from the gateway (captured during testing):

| Model | Streaming `finish_reason` |
| --- | --- |
| `deepseek-v4-pro` | `['stop']` ✓ |
| `glm-5` | `['stop']` ✓ |
| `muse-spark-1.2-contributor` | _(none sent)_ ✗ |

For `muse-spark-1.2-contributor`:
- streaming: no `finish_reason` chunk is ever sent
- non-streaming: `finish_reason` is `null`

The mapping path is:
1. `mapOpenAICompatibleFinishReason(null)` → `default` branch → `"other"`
2. streaming parser initializes `finishReason = { unified: "other" }`

=> every `muse-spark` response is reported as `finish="other"`.

## Expected behavior

A missing / `null` / `undefined` `finish_reason` should be treated as a normal completion (`"stop"`), not as `"other"`. Models that do send a proper `finish_reason` (e.g. `deepseek-v4-pro`, `glm-5`) already override this and are unaffected.

## Proposed fix

- `packages/opencode/src/provider/sdk/copilot/chat/map-openai-compatible-finish-reason.ts`: add a `null`/`undefined` guard before the `switch` that returns `"stop"`.
- `packages/opencode/src/provider/sdk/copilot/chat/openai-compatible-chat-language-model.ts`: change the streaming initial `finishReason.unified` default from `"other"` to `"stop"`.

A fix branch `fix/null-finish-reason-graceful` with tests will be opened as a PR referencing this issue.

anandlo:
## Summary

Some gateways served by the OpenCode Go gateway (`opencode.ai/zen/go/v1`) do **not** emit a `finish_reason` in streaming responses (and return `null` in non-streaming). The mapping then fell through to `default: return "other"`, which made `classify.ts` flag every such response as **degraded** / **think-only** even though the response was delivered normally.

Affected model: `muse-spark-1.2-contributor`.

### Evidence (live gateway streaming captures)

| Model | Streaming `finish_reason` |
| --- | --- |
| `deepseek-v4-pro` | `['stop']` ✓ |
| `glm-5` | `['stop']` ✓ |
| `muse-spark-1.2-contributor` | _(none sent)_ ✗ |

## Changes

- `map-openai-compatible-finish-reason.ts`: add a `null`/`undefined` guard before the `switch` that returns `"stop"`.
- `openai-compatible-chat-language-model.ts`: change the streaming initial `finishReason.unified` default from `"other"` to `"stop"` (models that send a proper `finish_reason` still override it).
- Added `test/provider/map-openai-compatible-finish-reason.test.ts` covering `null`/`undefined` plus regression cases (`stop`, `length`, `tool_calls`, `other`, `content_filter`, `""`).

## Verification

- `bun typecheck` from `packages/opencode/` passes.
- `bun test test/provider` passes: 482 pass / 0 fail (8 new tests included).

Fixes #2173

## Test plan

- [ ] Stream `muse-spark-1.2-contributor` and confirm `finish` is reported as `stop`, not `other`.
- [ ] Confirm `deepseek-v4-pro` / `glm-5` still report `stop` as before (no regression).

### DIFF
--- a/packages/opencode/src/provider/sdk/copilot/chat/map-openai-compatible-finish-reason.ts
+++ b/packages/opencode/src/provider/sdk/copilot/chat/map-openai-compatible-finish-reason.ts
@@ -3,6 +3,12 @@ import type { LanguageModelV3FinishReason } from "@ai-sdk/provider"
 export function mapOpenAICompatibleFinishReason(
   finishReason: string | null | undefined,
 ): LanguageModelV3FinishReason["unified"] {
+  // Null/undefined finish_reason indicates the gateway completed without an
+  // explicit signal (e.g. muse-spark via zen/go). Treat as "stop" since the
+  // response was delivered normally — defaulting to "other" would cause
+  // classify.ts to flag every such response as degraded.
+  if (finishReason == null) return "stop"
+
   switch (finishReason) {
     case "stop":
       return "stop"
--- a/packages/opencode/src/provider/sdk/copilot/chat/openai-compatible-chat-language-model.ts
+++ b/packages/opencode/src/provider/sdk/copilot/chat/openai-compatible-chat-language-model.ts
@@ -649,6 +649,13 @@ export class OpenAICompatibleChatLanguageModel implements LanguageModelV3 {
           },
 
           flush(controller) {
+            // If no explicit finish_reason was received (gateway omits it) and
+            // tool calls are present, default to "tool-calls" so the agent loop
+            // continues for tool execution rather than terminating as "stop".
+            if (finishReason.raw == null && toolCalls.length > 0) {
+              finishReason = { unified: "tool-calls", raw: undefined }
+            }
+
             if (isActiveReasoning) {
               controller.enqueue({
                 type: "reasoning-end",
--- a/packages/opencode/src/provider/sdk/copilot/chat/openai-compatible-chat-language-model.ts
+++ b/packages/opencode/src/provider/sdk/copilot/chat/openai-compatible-chat-language-model.ts
@@ -274,7 +274,12 @@ export class OpenAICompatibleChatLanguageModel implements LanguageModelV3 {
     return {
       content,
       finishReason: {
-        unified: mapOpenAICompatibleFinishReason(choice.finish_reason),
+        // Mirror the streaming flush() fallback: gateways that omit
+        // finish_reason but return tool calls ended in a tool call.
+        unified:
+          choice.finish_reason == null && choice.message.tool_calls != null && choice.message.tool_calls.length > 0
+            ? "tool-calls"
+            : mapOpenAICompatibleFinishReason(choice.finish_reason),
         raw: choice.finish_reason ?? undefined,
       },
       usage: {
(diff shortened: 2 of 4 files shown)


EsanRAHIMI (commented):
The missing-finish-reason fallback should account for tool calls before defaulting to `stop`. This gateway is described as omitting `finish_reason` entirely, so a streamed response containing `delta.tool_calls` will also keep the new initial `unified: "stop"` value. The transform still emits the tool calls in `flush()`, but the final finish part now says the model stopped normally rather than `tool-calls`; consumers that use the finish reason to decide whether to continue the agent/tool loop can terminate early. Could `flush()` set the fallback to `tool-calls` when `toolCalls.length > 0` (and no explicit reason was received), with a streaming regression test for a tool-call response that omits `finish_reason`?

anandlo:
The CI failure in `unit (shard 2/4)` is a **pre-existing issue on main**, not caused by this PR.

**Evidence:**
- Main branch CI (commit `fff7cffb`) also fails with the same MCP tests: `MCP structuredContent`, `exec can call a catalogued MCP tool`, etc.
- These MCP tests are unrelated to my changes in `map-openai-compatible-finish-reason.ts` and `openai-compatible-chat-language-model.ts`.
- All 4 tests specific to this PR pass locally: 20/20 in the copilot-chat-model and finish-reason test suites.

The MCP test failures appear to be a broken CI issue on main that predates this PR.

anandlo:
**Self-audit of this PR found and fixed 3 issues** (commit `48e50043`):

**1. Critical — flush() could clobber an `error` finish.**
The error path sets `{unified:"error", raw:undefined}` — its raw is also null, so my previous override condition (`raw == null && toolCalls.length > 0`) matched it too and rewrote error → tool-calls whenever tool deltas preceded a failed chunk. Now gated on `unified === "stop"`, which only the pristine default can have (any explicit reason sets raw; error sets unified). Regression test added: unparseable chunk + finish_reason-less tool calls must stay `error`.

**2. Important — non-streaming path had the same gap the review flagged for streaming.**
`doGenerate` with `finish_reason: null` + tool_calls reported `"stop"`. It now reports `"tool-calls"`, mirroring the flush() semantics.

**3. Coverage gap — the primary real-world case had no test.**
muse-spark's most common shape is a plain text stream with no finish_reason at all. Added a fixture asserting that surfaces as a clean `{unified: "stop", raw: undefined}`.

**Red/green verified:** the error-clobber test fails against the old condition and passes with the gate.

**Downstream safety checked:** classify.ts is the only consumer distinguishing "other" from "stop" (degraded warning only — both paths are terminal finals), and AI-SDK multi-step consumers key off "tool-calls", which both paths now report correctly.

Suite: 510 pass / 0 fail, typecheck clean.

PR #2195, opened 2026-08-21
fix(session): avoid _- in session IDs

wqymi:
## Summary
- Keep the 64-bit timestamp payload for session identifiers.
- Omit the internal sort marker from user-facing session IDs so resume commands never contain ses_-.
- Preserve v2 sort markers for other identifier types and add regression coverage.

## Tests
- bun test test/id/id.test.ts
- bun run typecheck
- Pre-push bun turbo typecheck (12/12 tasks successful)

### DIFF
--- a/packages/opencode/src/id/id.ts
+++ b/packages/opencode/src/id/id.ts
@@ -82,6 +82,13 @@ export function create(prefix: string, direction: "descending" | "ascending", ti
   }
 
   const mark = direction === "descending" ? "-" : "g"
+  // Session IDs are copied into the resume command. Session lists primarily
+  // order by update time, so keep this user-facing payload hexadecimal instead
+  // of exposing the v2 sort marker.
+  if (prefix === prefixes.session) {
+    return prefix + "_" + timeBytes.toString("hex") + randomBase62(LENGTH - TIME_BYTES * 2)
+  }
+
   return prefix + "_" + mark + timeBytes.toString("hex") + randomBase62(LENGTH - 1 - TIME_BYTES * 2)
 }
 
--- a/packages/opencode/test/id/id.test.ts
+++ b/packages/opencode/test/id/id.test.ts
@@ -29,10 +29,15 @@ describe("Identifier", () => {
   })
 
   test("descending v2 starts with - so it sorts before every v1 id", () => {
-    const next = Identifier.descending("session")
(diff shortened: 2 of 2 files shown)


packages/opencode/src/id/id.ts
```diff
@@ -82,6 +82,13 @@ export function create(prefix: string, direction: "descending" | "ascending", ti
   }
 
   const mark = direction === "descending" ? "-" : "g"
+  // Session IDs are copied into the resume command. Session lists primarily
```
MiMoHardFather: **[suggestion]** The new comment narrates the product rationale (resume copy-paste, list sort order, “hexadecimal instead of the v2 sort marker”) rather than a short WHY. The special case is non-obvious, so a comment belongs here, but this one restates the next three lines and embeds design history.

**Suggestion:** Collapse to one line, e.g. `// Session lists order by update time; omit the v2 marker so --session IDs stay hex.`

packages/opencode/test/id/id.test.ts
```diff
@@ -29,10 +29,15 @@ describe("Identifier", () => {
...
     expect(next[4]).toBe("-")
-    expect(next < v1("ses", "000000000000")).toBe(true)
-    expect(next < v1("ses", "ffffffffffff")).toBe(true)
+    expect(next < v1("msg", "000000000000")).toBe(true)
+    expect(next < v1("msg", "ffffffffffff")).toBe(true)
+  })
+
+  test("session IDs keep a hexadecimal payload without a sort marker", () => {
+    const next = Identifier.descending("session")
+    expect(next).toMatch(/^ses_[0-9a-f]{16}[0-9A-Za-z]{10}$/)
```
MiMoHardFather: **[suggestion]** The new test only asserts the generated session format (`/^ses_[0-9a-f]{16}[0-9A-Za-z]{10}$/`). Nothing locks in that `Identifier.schema("session")` still accepts historical v1 (`ses_` + 12 hex) and v2 (`ses_-` + 16 hex) IDs. A later change that copies this regex into `schema()` would break `--session` / `Session.get` for existing rows even though this PR itself does not.

**Suggestion:** Add schema assertions for a v1 session ID and a `ses_-…` v2 ID next to the generation check (mirroring the existing “schema accepts both v1 and v2” message test).

PR #2245, opened 2026-08-25
fix(cron): route fires to the creating session and add cron list --all

### LINKED ISSUE #2198: cron: cross-session fire leakage + list invisibility for non-durable jobs
Reproduced 2026-08-21/22 with two concurrent sessions in one process (non-durable `*/10` job `201577bd` created by session A; every fire injected into session B; from B, `cron list` empty under all filters while `get`/`delete` succeeded).

**Root causes**
1. `scheduler.ts:start()` is a per-process singleton (`if (rt) return`); `onFire` closes over the first mounter's sessionID and never consults `task.createdBySessionId` (recorded at scheduler.ts:298 but unused for dispatch).
2. Verb scopes disagree — `tool/cron.ts:381` hard-codes `list` to the caller while `findById` only filters when `--session` is explicitly passed.

**Fix in PR to follow**: route fires via `fireTargetSessionId(task, fallback)`; add `list --all`.

rekty:
{
  "title": "fix(cron): route fires to the creating session and add cron list --all",
  "head": "rekty:fix/cron-session-dispatch",
  "base": "main",
  "body": "Closes #2198\n\nTwo related bugs when multiple sessions share one process:\n\n## 1. Cross-session fire leakage\n\nThe scheduler is a per-process singleton (`if (rt) return`). The first session to mount it owns the `onFire` callback, and every subsequent session silently inherits it — even tasks created by session B land in session A's conversation.\n\n`fireTargetSessionId(task, fallback)` consults `task.createdBySessionId` (which `add()` has been recording since scheduler.ts:298 but nobody read) and falls back to the mounting session only for legacy tasks that lack the field.\n\n## 2. `cron list` scope hides other sessions' jobs\n\n`cron list` defaulted to the caller's session (`session_id: ctx.sessionID`), while `get`/`delete` only filtered on `--session` — so a job id the user could retrieve and cancel was invisible to `cron list`. Adding `--all` to `listOperation` tells the scheduler to skip the session filter entirely. When `all` is requested the owner is surfaced in the output so the caller can target `--session` correctly.\n\n## Changes\n\n| file | what |\n|---|---|\n| `cron/cron-task.ts` | `fireTargetSessionId(task, fallback)` helper |\n| `cron/scheduler.ts` | `ListFilter.all`; `list()` skips session filter when set |\n| `session/cron-bridge.ts` | `onFire` dispatches via `fireTargetSessionId` |\n| `tool/cron.ts` | `--all` on list (JSON + shell); owner shown in output |\n| `tool/cron.txt` / `cron.shell.txt` | document `all` |"
}

### DIFF
--- a/packages/opencode/src/cron/cron-task.ts
+++ b/packages/opencode/src/cron/cron-task.ts
@@ -97,6 +97,17 @@ export const getSessionCronTasks = (): CronTask[] => [...SESSION_STORE.values()]
 export const removeSessionCronTasks = (ids: string[]) =>
   ids.forEach((id) => SESSION_STORE.delete(id))
 
+/**
+ * Which session a fire should be injected into.
+ *
+ * The scheduler is a per-process singleton: the first session to mount it owns
+ * the onFire callback, but tasks record who created them. Routing by the
+ * creator keeps session B's recurring job from leaking into session A's
+ * conversation when A happened to mount the scheduler first (#2198).
+ */
+export const fireTargetSessionId = (task: CronTask, fallback: string): string =>
+  task.createdBySessionId ?? fallback
+
 export const findMissedTasks = (tasks: CronTask[], now: number): CronTask[] =>
   tasks.filter((t) => {
     if (t.recurring) return false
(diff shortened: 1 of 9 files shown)


rekty:
cc @MiMoHardFather @wqymi @yanyihan-xiaomi — could one of you approve the CI workflow runs for this fork PR? The lint/test/typecheck checks are waiting on approval. Thanks!

PR #2248, opened 2026-08-25
fix(config): tolerate bare permission words in MIMOCODE_PERMISSION env var

### LINKED ISSUE #2170: [bug] Startup crash: JSON Parse error: Unexpected identifier "allow" when MIMOCODE_PERMISSION is set as a bare word
## Environment
- Windows 11, MiMoCode v0.1.12 (Bun build)
- Reproduced on every subcommand: `mimo`, `mimo debug config`, `mimo session list`

## Symptom
Every `mimo` invocation crashes at startup with:

``
JSON Parse error: Unexpected identifier "allow"
``

and the TUI shows `tui bootstrap failed`. This happens regardless of config/data state: after wiping all of `~/.config/mimocode`, `~/.local/share/mimocode` and `~/.cache/mimocode`, the exact same crash occurs.

## Root cause
The **environment variable** `MIMOCODE_PERMISSION` was set (persisted in the Windows registry, `HKCU:Environment`) to the bare word `allow`:

``
setx MIMOCODE_PERMISSION allow
``

MiMoCode parses this variable as JSON at startup. The documented correct value (see permissions.md: *(env)* `MIMOCODE_PERMISSION='"allow"'`) is the **quoted string** `"allow"`. A bare `allow` is invalid JSON → `Unexpected identifier "allow"` → crash on every launch.

Because environment variables are snapshotted per-process, the failure also persists in already-open terminals even after fixing the registry value — the only recovery is: delete the registry value **and** open a fresh terminal.

## Suggested fix
1. Validate the value and surface a clear, actionable error (e.g. *"MIMOCODE_PERMISSION must be valid JSON; got bare 'allow' — did you mean \"allow\"?"*) instead of a raw `JSON Parse error`.
2. Optionally tolerate the bare words `allow` / `ask` / `deny` (the only documented action values) when JSON.parse fails.
3. A single bad env var should not fatally break *every* subcommand including headless ones (`debug`, `session list`, `providers`) — a warning would be enough.

Crash backtrace:
``
SyntaxError: JSON Parse error: Unexpected identifier "allow"
  at <parse> (:0)
  at parse (unknown)
  at <anonymous> (B:/~BUN/root/chunk-3egw8k9j.js:536:5321)
  ...
  at processTicksAndRejections (native:7:39) fatal
``

rekty:
## Summary

Wraps `JSON.parse(MIMOCODE_PERMISSION)` in a try/catch so bare words like `allow` (a common mistake from `setx MIMOCODE_PERMISSION allow`) no longer crash every subcommand.

## Problem

Setting `MIMOCODE_PERMISSION=allow` as a bare word (without JSON quotes) crashes MiMoCode at startup on every subcommand with `JSON Parse error: Unexpected identifier "allow"` — including headless ones like `debug` and `session list`. Recovery requires manually editing the Windows registry.

## Fix

1. Tolerate bare words `allow`, `ask`, `deny` → treat as `{ "*": value }` with a warning
2. Other bare words → clear error message with hint on correct syntax
3. Valid JSON values → work exactly as before (no behavior change)

**Changed:** 1 file, +23/−1 (`packages/opencode/src/config/config.ts`)

Closes #2170

CC @MiMoHardFather @wqymi @yanyihan-xiaomi — please approve CI runs for first-time contributor. 🙏

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -971,7 +971,29 @@ export const layer = Layer.effect(
         }
 
         if (Flag.MIMOCODE_PERMISSION) {
-          result.permission = mergeDeep(result.permission ?? {}, JSON.parse(Flag.MIMOCODE_PERMISSION))
+          const raw = Flag.MIMOCODE_PERMISSION
+          let parsed: Record<string, unknown>
+          try {
+            parsed = JSON.parse(raw)
+          } catch {
+            // Tolerate bare action words like "allow" / "ask" / "deny" which
+            // are the only documented permission actions.  A bare word is not
+            // valid JSON, but it is a common mistake (e.g. `setx MIMOCODE_PERMISSION allow`).
+            const bare = raw.trim().toLowerCase()
+            if (bare === "allow" || bare === "ask" || bare === "deny") {
+              parsed = { "*": bare } as ConfigPermission.Info
+              console.warn(
+                `[config] MIMOCODE_PERMISSION="${raw}" is not valid JSON — treating as { "*": "${bare}" }. ` +
+                `Use JSON syntax for future: MIMOCODE_PERMISSION='"${bare}"'`,
+              )
+            } else {
+              throw new Error(
+                `MIMOCODE_PERMISSION must be valid JSON (e.g. '"allow"') — got bare word '${raw}'. ` +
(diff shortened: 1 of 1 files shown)


rekty:
Hi @MiMoHardFather @wqymi @yanyihan-xiaomi! This PR fixes a startup crash when MIMOCODE_PERMISSION is set to a bare word. Could you please approve the CI workflow runs? Thank you! 🙏

PR #2261, opened 2026-08-26
fix(actor): drop the unimplemented spawn/run resume and make send→wait reliable

yanyihan-xiaomi:
## Summary

Two defects behind one user-visible symptom: a model trying to continue with a subagent it already spawned got a brand-new empty one, and when it used the flow that *does* work (`send`) the follow-up result was unreliable.

**`actor_id` on spawn/run was a promise with no implementation.** `src/tool/actor.ts` accepted it and described it as "resume the specified prior actor session instead of creating a new one", but `execute` never read `op.actor_id` — `Actor.spawnSubagent` unconditionally calls `allocateActorID` (max+1). An old test pinned the real behaviour (`// v9: actor_id in run action is ignored`) while its title claimed it resumed. Per maintainer decision this shrinks the contract rather than implementing resume: the argument is gone from the JSON schema, the shell `--actor` flag, `ActorShellArgs` and `recoverActorArgs`, and both description files now document the `send` → `wait` follow-up flow. `--actor` on the shell path fails with a named error pointing at the replacement instead of a bare arity mismatch.

**An inbox-woken turn bypassed the actor status machine.** `actor/turn.ts:runTurn` is the only writer of actor `running`/`idle`+outcome and the only publisher of `ActorStatusChanged`, and it wrapped only the spawn turn (plus preStop/gate/postStop) inside `forkWork`. The `actor send` wake path is `Inbox.send → SessionPrompt.loop → runLoop` and never touched actor status, so for the whole woken turn the registry row still read `idle+success` from the **previous** turn — precisely what `ActorWaiter`'s fast path treats as finished, at which point it answers from the slice's last assistant message. A `wait` issued while the woken turn was still in flight therefore returned the *previous* turn's answer. It was a race, not a hard failure: if the woken turn had already finished, the same fast path returned the correct new answer, which is why the flow appeared to work much of the time.

The fix has two halves. `Inbox.send` calls the new `ActorRegistry.markPending` — a `WHERE status='idle'` guarded update — before the fork-and-forget wake, closing the window where `send` returns before the woken turn marks itself running. And `SessionPrompt.loop` wraps a woken turn in `runTurn`. Two constraints on that wrapper, both found by review:

- It is gated on the inbox wake flag rather than on a non-`main` `agentID`, because `SessionPrompt.prompt` also ends in `loop()` and `forkWork` already wraps that call. Nesting them let the inner `runTurn` publish `idle+success` before `runAgentLoop` read the assistant error and failed the outer.
- It raises a settled assistant error itself, because `runLoop` returns one as a **success value** (it returns the errored message and only branches on `finalIsError` to notify the parent). Without that, a failed woken turn was recorded — and reported to a waiting sender — as `idle+success`.

`markPending` also advances `last_activity_time`: `deriveLiveness` measures a `pending` row against that timestamp, so a subagent idle longer than the stall bound would have been announced as stalled the instant a message queued it. It publishes only on a real transition (via `RETURNING`), since `main` rows sit at `pending` permanently and receive every subagent notification.

## Subagent completion semantics are unchanged

Verified explicitly, since preStop/gate/postStop each run extra turns and the "did it finish" / "which turn's text is returned" answers are load-bearing. All three of `forkWork`'s `runTurn` call sites are unaffected by the new wrapper — they reach `loop` without the wake flag. `test/actor/poststop-progress-write-permission.repro.test.ts` now pins the attribution across a spawn that ran a gate re-entry *and* a postStop turn: the outcome carries the gate turn's text plus the gate's reconciliation suffix and its `partial` downgrade, and not the postStop turn's text. **That assertion passes against unmodified `src/` as well**, which is what demonstrates the semantics were not moved.

## Verification

- `bun typecheck` (packages/opencode) — clean.
- `bun test test/actor/ test/inbox/ test/tool/actor.test.ts test/tool/actor.shell.test.ts` — 262 pass / 2 skip / 0 fail.
- `bun test test/session/` — 964 pass / 25 skip / 1 todo / 0 fail.
- Every new regression test was confirmed to fail with its corresponding fix reverted: the woken-turn state trajectory, the ungated wrapper (`lastOutcome` became `success`), the unraised assistant error (same), and the liveness write (read `idle`).

## Known gaps, deliberately not addressed

- `resume`/`resumeBackground` run actor turns without the status machine.
- When the registry reads `idle` but the Runner is still busy, `ensureRunning` returns the in-flight Deferred without executing work, so the row can sit at `pending` until the next turn writes it.
- `ActorRegistry.defaultLayer` and `ActorWaiter.defaultLayer` each build a private Bus, and Effect 4 beta layer memoization does not span `Layer.mergeAll` branches without a shared MemoMap (production shares one via `ManagedRuntime.make(AppLayer, { memoMap })`). This is why the end-to-end test asserts the registry trajectory while the waiter-level assertions live in `waiter.test.ts`, and it is very likely the real cause of the long-skipped slow-path test there — that test passes when run alone.

### DIFF
--- a/packages/opencode/src/actor/registry.ts
+++ b/packages/opencode/src/actor/registry.ts
@@ -75,6 +75,11 @@ export interface Interface {
       lastError?: string | undefined
     },
   ) => Effect.Effect<void>
+  // Flip an IDLE actor back to `pending` for the inbox wake path: a queued message
+  // means it has work again, and a row left `idle` lets ActorWaiter resolve on the
+  // PREVIOUS turn's outcome before the woken turn marks itself running. A row that
+  // is not idle is left alone.
+  readonly markPending: (sessionID: SessionID, actorID: string) => Effect.Effect<void>
   readonly updateTurn: (sessionID: SessionID, actorID: string) => Effect.Effect<void>
   readonly updateAgent: (sessionID: SessionID, actorID: string, agent: string) => Effect.Effect<void>
   readonly get: (sessionID: SessionID, actorID: string) => Effect.Effect<Actor | undefined>
@@ -169,6 +174,35 @@ export const layer: Layer.Layer<Service, never, Bus.Service> = Layer.effect(
       return fromRow(row)
     })
 
+    // Re-read after a status write so the event payload reflects committed row
+    // values rather than the sparse patch, then publish. A row that vanished
+    // between UPDATE and SELECT publishes nothing — a dropped event beats a
+    // misleading one.
+    const publishStatus = Effect.fn("ActorRegistry.publishStatus")(function* (
+      sessionID: SessionID,
+      actorID: string,
+    ) {
+      const row = yield* Effect.sync(() =>
+        Database.use((db) =>
+          db
+            .select()
+            .from(ActorRegistryTable)
+            .where(and(eq(ActorRegistryTable.session_id, sessionID), eq(ActorRegistryTable.actor_id, actorID)))
+            .get(),
+        ),
+      )
+      if (!row) return
+      yield* bus.publish(Events.ActorStatusChanged, {
+        sessionID,
+        actorID,
+        status: row.status,
+        ...(row.last_outcome ? { lastOutcome: row.last_outcome } : {}),
+        turnCount: row.turn_count,
+        lastTurnTime: row.last_turn_time,
+        ...(row.last_error ? { error: row.last_error } : {}),
+      })
+    })
+
     const updateStatus = Effect.fn("ActorRegistry.updateStatus")(function* (
       sessionID: SessionID,
       actorID: string,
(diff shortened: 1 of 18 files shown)


yanyihan-xiaomi:
Closing in favour of a minimal replacement.

This PR grew past its own remit. It set out to remove the `actor_id` argument that spawn/run never implemented, and ended up also rewriting the actor status machine so that `send` -> `wait` would be reliable. That second half kept failing review: four rounds found six defects in it, and the last audit showed the central promise still was not met — `wait` on a *running* receiver continues to answer with the previous turn's text, because the fix only ever covered an idle one.

The `wait` staleness is a pre-existing defect, not a regression from this branch: on unmodified main, a send followed by wait already returns the previous turn's answer whenever the woken turn has not finished yet. It deserves its own change, designed around the right signal (whether the receiver still has an undrained inbox row) rather than around forcing the registry row out of `idle`.

Replacing this with a minimal PR that only removes the unimplemented argument and points models at `send`, documenting the `wait` mechanism as a known, separate defect.

PR #2262, opened 2026-08-26
fix(actor): drop the unimplemented spawn/run resume argument

yanyihan-xiaomi:
## Summary

`actor_id` on `spawn`/`run` was documented as resuming a prior subagent — the description said the call "continues with its prior messages and tool outputs" — but `execute` never read it. `Actor.spawnSubagent` unconditionally calls `allocateActorID`, so a model trying to continue with a subagent it already had silently received a brand-new one with an empty context. The behaviour was even pinned by a test whose title claimed the opposite (`// v9: actor_id in run action is ignored — always creates new` under *"execute resumes an existing task session from actor_id"*).

This removes the argument rather than implementing resume. `send` already appends to a subagent's own history and wakes it for another turn, so the descriptions now point at that instead.

Removal covers every surface: the JSON schema for both actions, the shell `--actor` flag, the `ActorShellArgs` type, and the recovery path. The recovery path is the subtle one — `recoverActorArgs` lifts a bare `{subagent_type, description, prompt}` into a valid operation, and its whitelist would have silently *dropped* a top-level `actor_id`, recovering the call into a valid spawn and reproducing exactly the silent failure being removed. It now carries `actor_id` through so the strict schema rejects the call instead. That path is the one a model confused enough to still pass `actor_id` is most likely to take.

## Known defect, not addressed here (pre-existing, not a regression)

While tracing the above I found that collecting a woken subagent's answer is unreliable, and I want it recorded rather than quietly bundled into this change.

`actor/turn.ts:runTurn` is the only writer of actor `running`/`idle`+outcome and the only publisher of `ActorStatusChanged`, and it wraps only the turns `forkWork` drives. The `actor send` wake path — `Inbox.send` → `SessionPrompt.loop` → `runLoop` — never touches actor status, so for the whole woken turn the registry row still reads `idle`+`success` from the *previous* turn. `ActorWaiter.isWaitResolving` treats any `idle` ephemeral row as finished (`src/actor/waiter.ts:35-38`) and then answers from the slice's last assistant message, so a `wait` issued while the woken turn is still in flight returns the **previous** turn's text.

This is not introduced by this PR: on unmodified `main`, a `send` followed by a `wait` already returns the previous answer whenever the woken turn has not finished yet. It is a race in the idle case and a much wider window when the receiver is mid-turn, because `SessionRunState.ensureRunning` hands back a busy runner's Deferred *discarding* the work it was given (`src/effect/runner.ts:118`), so nothing re-stamps the row for the duration of a full model call.

I attempted the fix in a previous version of this branch and withdrew it: forcing the row out of `idle` only covers an idle receiver, and it disturbed several consumers of actor status (group barriers, TUI tool-part animation, cancelled-outcome bookkeeping). The right signal is whether the receiver still has an undrained inbox row, which is authoritative because `drain` deletes it — that belongs in its own change, scoped to `ActorWaiter`.

Until then, the tool descriptions steer models to the completion notification rather than to `wait` for a woken turn's result, which is the behaviour that does hold today.

## Verification

- `bun typecheck` (packages/opencode) — clean.
- `bun test test/tool/actor.test.ts test/tool/actor.shell.test.ts test/tool/actor-recover.test.ts` — 67 pass / 0 fail.

Test changes: the mis-titled resume test is replaced by strict schema rejections of `actor_id` on both actions; `--actor` on the shell path now fails as a named unknown flag pointing at `send`, instead of a bare arity mismatch from the positionals; and `recoverActorArgs` gains a test pinning that a top-level `actor_id` is carried through to be rejected rather than dropped.

### DIFF
--- a/packages/opencode/src/tool/actor.shell.txt
+++ b/packages/opencode/src/tool/actor.shell.txt
@@ -21,13 +21,13 @@ appends them at request time per-agent).
 #   - $vars are preserved as literal text (no expansion)
 
 # DEFAULT — return actor_id immediately, run in the background, stay responsive:
-    actor spawn <subagent_type> "<description>" "<prompt>" [--model <ref>] [--actor <id>] [--command <cmd>] [--context none|state|full] [--output-schema '<json>']
+    actor spawn <subagent_type> "<description>" "<prompt>" [--model <ref>] [--command <cmd>] [--context none|state|full] [--output-schema '<json>']
     # bind it to one of your `task` tool tasks (TID from the `task` tool, e.g. T4):
     actor spawn <subagent_type> "<description>" "<prompt>" --task <TID>
 
 # EXCEPTION — block the whole conversation until done, return the final message inline.
 # Only for a tiny, fast lookup whose answer gates your very next decision this turn:
-    actor run <subagent_type> "<description>" "<prompt>" [--model <ref>] [--actor <id>] [--timeout <ms>] [--command <cmd>] [--context none|state|full] [--output-schema '<json>']
+    actor run <subagent_type> "<description>" "<prompt>" [--model <ref>] [--timeout <ms>] [--command <cmd>] [--context none|state|full] [--output-schema '<json>']
     # bind it to one of your `task` tool tasks (TID from the `task` tool, e.g. T4):
     actor run <subagent_type> "<description>" "<prompt>" --task <TID>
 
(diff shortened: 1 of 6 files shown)


yanyihan-xiaomi:
CI status: lint, typecheck, and unit shards 1–3 are green. The only red is `unit (shard 4/4)`, failing on `test/session/auto-overflow-writer-first.test.ts` → "a completed high-usage turn is rebuilt exactly once".

That failure is **pre-existing on `main`** and unrelated to this PR:
- The same case fails on `main`'s own CI (run 33036779783, at the current `main` HEAD).
- This PR touches only `src/tool/actor.{ts,txt}`, `src/tool/actor.shell.txt`, and the three `test/tool/actor*` files — nothing under `src/session/` or that test.
- Reproduced locally: the case passes in isolation and fails intermittently in the full shard, i.e. it is a flaky/broken session-compaction test, not a regression here.

Branch is rebased onto the latest `main`. Ready for a **squash** merge; the shard-4 red should not block it. Leaving the merge to a human rather than using `--admin` to bypass the check.

PR #2267, opened 2026-08-27
fix(session): include MCP tools in the fork/checkpoint request prefix

yanyihan-xiaomi:
## Summary

`buildLLMRequestPrefix` (the single source that builds a fork's — and a `fork:false` checkpoint-writer's — LLM request prefix) assembled its tool set from **registry built-ins only**, while the parent's real request merges MCP tools in `SessionPrompt.resolveTools`. So the moment any MCP server is configured, `ForkContext.tools` diverged from the parent on the **first turn**:

- The tool-list prefix cache misses immediately (the tools array differs).
- A `fork:false` checkpoint-writer sends this schema directly, so it runs **without the MCP tools** the parent has.

The `ForkContext.tools` JSDoc even documented the (wrong) assumption behind this — "both paths call `registry.tools` … so schema parity holds" — which silently omitted MCP.

This was surfaced by dumping a checkpoint-writer request vs. the equivalent non-checkpointed request: the rebuilt request was missing the two `exa_*` MCP tools that the parent carried.

## What changed

`buildLLMRequestPrefix` now reproduces the parent's **provider-visible** MCP set, matching both toolset modes:

- **Non-GPT toolset** — append every connected, executable, non-name-clashing MCP tool, minus those the merged session/agent permission denies or the turn toggles off (`user.tools`). The agent `toolAllowlist` does **not** prune the wire set, so it is deliberately not applied.
- **GPT/Codex toolset** — MCP tools never enter `activeTools` (they are reached through `mcp-tool-search`), so the parent emits **zero** MCP definitions on the wire; the prefix skips the MCP append entirely to match.

The per-turn toggle is read from the last user message's `tools` map — the same input `LLM.resolveTools` uses — so the parent runLoop and the checkpoint-capture closure share one source and cannot drift.

Also corrects the `ForkContext.tools` JSDoc (`spawn.ts`) and the parent runLoop comment (`prompt.ts`) to describe the real two-layer wire tool set (`SessionPrompt.resolveTools` admits all connected MCP → the AI SDK filters by `activeTools`) instead of the old "both call registry.tools" claim.

## Why the two-mode split

The wire tool set the AI SDK sends is `tools` filtered by `activeTools` (`prepareToolsAndToolChoice`). `SessionPrompt.resolveTools` puts every connected MCP tool into the `tools` map (`prompt.ts:1847`), but under the GPT toolset those tools never join `activeTools` (`prompt.ts:1705` `&& !useGPTTools`, `:1918`), so the parent emits none. Listing them in the prefix would re-break parity — hence the `usesGPTToolset` guard.

## Verification

- `bun typecheck` (packages/opencode) — **0 errors**.
- `bun test test/session/llm-request-prefix.test.ts` — **4 pass, 2 skip, 0 fail**. New tests: MCP tools included (non-GPT), permission-deny excluded, per-turn toggle excluded, GPT/Codex toolset omits MCP.
- `bun test test/session/checkpoint-fork-mode.test.ts test/agent/agent.test.ts test/session/replace-agent-subagent.test.ts test/inbox/fork-agent-compat.test.ts` — **all pass**, no regressions.
- Note: `test/session/auto-overflow-writer-first.test.ts "rebuilt exactly once"` fails on `origin/main` unmodified (verified via `git stash`); it is a pre-existing baseline failure being fixed separately and is unrelated to this change.

### DIFF
--- a/packages/opencode/src/actor/spawn.ts
+++ b/packages/opencode/src/actor/spawn.ts
@@ -73,13 +73,24 @@ export interface ForkContext {
    * Tool schema as parent would emit at watermark, captured for invariant
    * verification only. NOT consumed by fork's runLoop for the actual LLM
    * request — that uses `resolveTools(forkAgent)` for executable tools with
-   * dispatch closures. Schema parity is currently enforced because
-   * checkpoint-writer has no `toolAllowlist` (Task 2.6); both paths call
-   * `registry.tools` with equivalent agent inputs and produce identical
-   * schemas. If `toolAllowlist` is ever re-added, this field would still
-   * snapshot parent's schema while the runtime tools would diverge, silently
-   * breaking cache parity. Test guard: `test/agent/agent.test.ts` asserts
-   * `cp.toolAllowlist === undefined` for checkpoint-writer.
+   * dispatch closures. `buildLLMRequestPrefix` reproduces the parent's WIRE tool
+   * set: registry built-ins plus the connected MCP tools that actually reach the
+   * provider. The AI SDK filters wire tools by `activeTools`, so the MCP parity
+   * target is "connected MCP tools that end up in activeTools":
+   *   - Non-GPT toolset: every executable, non-name-clashing MCP tool minus those
+   *     the merged session/agent permission denies or the turn toggles off
+   *     (user.tools). The agent toolAllowlist does NOT prune the wire set, so it
+   *     is not applied.
+   *   - GPT/Codex toolset: MCP tools never enter activeTools (they are reached via
+   *     mcp-tool-search), so the parent emits ZERO MCP definitions and the prefix
+   *     omits them too.
+   * Before MCP was folded in here, this field carried built-ins alone, so the
+   * mere presence of an MCP server diverged the fork prefix from the parent on
+   * the first turn and broke tool-list cache parity (and stripped MCP from
+   * fork:false checkpoint-writer requests, which send this schema directly).
+   * Test guard: `test/session/llm-request-prefix.test.ts` covers MCP inclusion,
+   * deny/toggle pruning, and GPT-toolset omission; `test/agent/agent.test.ts`
+   * asserts `cp.toolAllowlist === undefined` for checkpoint-writer.
    */
   readonly tools: Record<string, AITool>
   /**
--- a/packages/opencode/src/session/llm-request-prefix.ts
+++ b/packages/opencode/src/session/llm-request-prefix.ts
@@ -1,13 +1,16 @@
 import { Effect } from "effect"
-import { tool, jsonSchema, type Tool as AITool } from "ai"
+import { tool, jsonSchema, asSchema, type Tool as AITool } from "ai"
 import z from "zod"
 import { MessageV2 } from "./message-v2"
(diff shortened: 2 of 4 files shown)


yanyihan-xiaomi:
Closing this PR: further investigation showed it does not fix the problem it claims to, and the field it touches has no runtime effect.

## Why this does not work

`ForkContext.tools` — the field this PR populates with MCP schemas — **has no consumers**. A repo-wide search finds it referenced only in comments; both fork modes (`fork:true` and `fork:false`) build their real request tools from `SessionPrompt.resolveTools`, as noted at `session/prompt.ts:4106`. The parent runLoop likewise discards the `tools` returned by `buildLLMRequestPrefix` and uses `resolveTools`' map (`prompt.ts:4371`). So this change makes a verification-only snapshot more accurate and alters no wire request.

## The actual cause of the symptom

The motivating evidence was a dumped checkpoint-writer request carrying 29 built-in tools but none of the 2 configured `exa_*` MCP tools, while the equivalent non-checkpointed request carried all 31. That gap is produced by the **checkpoint-writer actor tool whitelist**, not by the prefix builder:

- The writer is spawned with `tools: ["read", "write", "edit", "apply_patch", "glob", "grep", "task"]` (`session/checkpoint.ts:937`).
- In `resolveTools`, that whitelist feeds `searchable`, which gates MCP membership in `activeTools` (`prompt.ts:1697`, `:1705`).
- Built-in tools enter `activeTools` **without** any whitelist check (`prompt.ts:1670`).
- The AI SDK then filters the wire tool list by `activeTools` (`ai/dist/index.mjs:1789-1790`).

Net effect: all 29 built-ins survive, every MCP tool is dropped — exactly the dumped shape. The asymmetry between how built-ins and MCP tools are gated by the actor whitelist is the real issue, and fixing it means changing wire behaviour for spawned actors, which is out of scope here.

## Review findings that remain valid (for whoever picks this up)

A review bot flagged five gaps in this implementation relative to the parent's wire set. All are technically correct, and all are documentation-level in impact because of the unused-field situation above. Any future attempt at prefix/wire tool parity must reproduce the **full** `activeTools` membership condition, not just permission-deny and per-turn toggle:

1. `isMcpToolSearchEnabled` — with `MIMOCODE_EXPERIMENTAL_MCP_TOOL_SEARCH` (or the umbrella experimental flag) MCP goes through `mcp-tool-search` even on non-GPT models, so `activeTools` holds no MCP tools up front (only ones loaded later, `prompt.ts:1918`).
2. `model.capabilities.toolcall` — required at `prompt.ts:1705`.
3. Agent `toolAllowlist` — it *does* prune the wire set via `activeTools`; the JSDoc wording added in this PR claiming otherwise is wrong.
4. Actor whitelist (`actor.tools`) — participates in `searchable` (`prompt.ts:1697`) and is the mechanism behind the symptom above.
5. `toolRegistry.tools(...)` is called here without `apiModelID` / `family`, unlike `prompt.ts:1561-1567`, so the GPT-toolset determination can disagree between the two paths when only `api.id` contains `gpt`.

No behavioural regression is being reverted by closing this — the branch `wt/fix-fork-mcp-tools` is left in place for reference.

PR #2273, opened 2026-08-28
fix(permission): same-session subagents inherit parent grants

wqymi:
## Summary
- `decideAskRouting` only consulted `session.parentID`, which is empty on a root session.
- Actor spawn/run subagents share the parent session, so they skipped `inherit` and fail-closed on every ask unless instance `skip-all` was already on.
- Fall back to the current session id when resolving the inherit parent (`mode: subagent` only).
- Add unit coverage for same-session inherit and for skip-all inheritance of background subagents.
- Unit CI shards were position-based (`NR % n`); adding one test file reshuffled every later file and broke order-sensitive tests (`mcp/sampling-e2e`). Switch sharding to `cksum(path) % n`.
- Isolate `mcp/stdio-exit-observe.test.ts` from the shared bun process: it needs the real `ObservingStdioTransport`, which `lifecycle.test.ts` `mock.module`s process-wide.

## Scope
Full access still depends on instance `skip-all`. This PR does not change that path; it only restores inheritance for paths the parent already approved.

## Test plan
- [x] `test/agent/ask-routing.test.ts`
- [x] `test/permission/inherit.test.ts` (same-session always-grant + skip-all cases)
- [x] Local hash-shard partition: insert moves 0 existing files
- [x] `stdio-exit-observe` passes in isolation
- [x] CI unit shards 1–4 on this PR
- [x] Live TUI smoke: parent `Always allow` on `external_directory`, then same-session `general` spawn writes the same directory with no second ask and no hang

## Live verification
Drived the TUI from this branch (`bun dev`, throwaway workspace, no `skip-all`):

1. Parent `write` outside the worktree raised one `external_directory` ask for `/tmp/.../ext/*`.
2. User chose **Allow always**.
3. `actor spawn` a `general` subagent; it `write`d `/tmp/.../ext/sub.txt`.
4. Worker log shows exactly **one** `service=permission ... asking` for the whole run.
5. Subagent write completed in 8ms (`tool=write ... ok=true`); parent then `actor wait`ed and finished.

Before this change the same-session subagent skipped `inherit` and fail-closed, so `sub.txt` would not be created.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,15 +37,49 @@ jobs:
       # runtime-worktree.test.ts excluded from shards: its deadline-fired test
       # deadlocks when run after other tests (QuickJS WASM state leak).
       # It runs separately below in an isolated process.
+      #
+      # stdio-exit-observe.test.ts excluded from shards: it needs the real
+      # ObservingStdioTransport. lifecycle.test.ts mock.module's that path
+      # process-wide, so sharing a bun test process with it always breaks
+      # these tests. Same isolation pattern as runtime-worktree.
+      #
+      # Sharding is path-hash, NOT list position: `NR % n` on the sorted file
+      # list reassigns every later file when one is added/removed, which
+      # reshuffles process-wide mocks and breaks order-sensitive tests
+      # (e.g. mcp/sampling-e2e). hash(path) % n keeps each file's shard
+      # independent of its neighbors.
+      #
+      # Scope: this stabilizes shard membership and isolates the one pair that
+      # actually broke. It is NOT a general mock.module isolation scheme —
+      # other files that mock.module still share a bun test process with their
+      # shard neighbors. If a new mock lands on the same hash bucket as a test
+      # that needs the real implementation, isolate that file the same way as
+      # runtime-worktree / stdio-exit-observe rather than assuming hash-sharding
+      # alone is enough.
       - name: Run unit tests (shard ${{ matrix.shard }})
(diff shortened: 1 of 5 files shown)


wqymi:
### Live TUI verification (this branch)

Reproduced the same-session inherit path in a real TUI run of `b46a7573ce`:

| Step | Result |
|------|--------|
| Parent `write` outside worktree | one `external_directory` ask for `/tmp/.../ext/*` |
| User reply | **Allow always** |
| `actor spawn` `general` | spawned in the same session |
| Subagent `write` `/tmp/.../ext/sub.txt` | success, 8ms, no second ask, no hang |
| Permission asks in the whole run | **1** |
| Parent `actor wait` + final reply | completed |

Worker log excerpt (`dev-worker-9200`):

```
12:35:13 service=permission id=... permission=external_directory patterns=["/tmp/.../ext/*"] asking
12:35:38 POST /permission/.../reply
12:35:38 service=format file=/tmp/.../ext/parent.txt formatting
12:36:19 service=llm ... agent=general mode=subagent stream
12:36:28 service=session.prompt tool=write ... tool execute start
12:36:28 service=format file=/tmp/.../ext/sub.txt formatting
12:36:28 service=session.prompt tool=write ... durationMs=8 ok=true tool execute done
```

`asking` count for the entire run: **1**. The subagent write never entered the human ask path — it inherited the parent's published grant via the new `sessionID` fallback.

CI on this PR is green: lint, typecheck, unit shards 1–4.

PR #2295, opened 2026-08-31
fix(pi-102): defer overflow cleanup for in-flight tool calls

pyramidheadshark:
## Summary
- `SessionProcessor.cleanup()` unconditionally waited only 250ms for an in-flight
  tool call's Deferred before force-aborting it, regardless of why cleanup ran.
- Overflow is detected at `finish-step`, but a slow local tool (bash/MCP/file I/O)
  can still genuinely be executing past that point — the 250ms grace then forced
  it aborted while the real side effect kept running in the background, producing
  a torn transcript (aborted tool-part + still-running side effect) that later
  triggered a server_error.
- Additive fix: new `MIMOCODE_OVERFLOW_TOOLCALL_GRACE_MS` flag (default 120000ms),
  applied ONLY when cleanup runs because `ctx.needsOverflowHandling` is true.
  Every other exit reason (interrupt/error/blocked/normal) keeps the original
  unconditional 250ms untouched.

## Test plan
- [ ] `bun run typecheck` (CI) — 0 new errors
- [ ] `bun test packages/opencode/test/session/` (CI)
- [ ] Live repro of the PI-102 scenario (compaction fires mid active tool call) on a built binary — no more server_error/silent agent switch


pyramidheadshark:
Sorry — opened against the wrong repo by mistake (gh defaulted to the fork's upstream parent). Closing; the actual PR lives in my own fork.

PR #2296, opened 2026-08-31
fix(actor): auto-Resume once after a retryable settle (main-agent parity)

wqymi:
## Summary
Background actors already share the main agent's session-level retry ladder. When that budget exhausts on a transient API error, the assistant is a /recovery candidate — the same path desktop Resume continues — but runAgentLoop immediately hard-failed the actor.

This change auto-calls SessionPrompt.resume once on a retryable settle. If resume is unavailable, the original AssistantSettledError classification is preserved. Non-retryable failures unchanged. No spawn-layer retry budget.

Spec: docs/compose/spec/actor-resume-parity.md

## Verification
- packages/opencode bun test test/actor/spawn.test.ts → 32 pass
- typecheck clean

### DIFF
--- a/docs/compose/spec/actor-resume-parity.md
+++ b/docs/compose/spec/actor-resume-parity.md
@@ -1,0 +1,39 @@
+---
+feature: actor-resume-parity
+status: in-progress
+updated: 2026-08-31
+branch: fix/actor-resume-parity
+commits: # filled at delivery
+---
+
+# Actor resume parity with main agent
+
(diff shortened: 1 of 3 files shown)


wqymi:
Superseded: actor auto-Resume is not the desired strategy. Background actor failures should surface to the parent for re-dispatch, not auto-continue. The product issue to fix is desktop losing the running state while the main agent is still busy.

PR #2327, opened 2026-09-03
fix(session): send x-opencode-session on outbound requests

### LINKED ISSUE #2317: Send x-opencode-session header on API requests (required by OpenCode Go from 09/05)
Hi — we operate the OpenCode Go managed-inference API, and 280 of your users' orgs call it through MiMoCode (UA `mimocode/0.1.13 ai-sdk/provider-utils/4.0.23 runtime/bun/1.3.14`).

Starting 09/05, requests without an `x-opencode-session` header will error on our side — we need a stable per-conversation ID for routing/optimization. MiMoCode doesn't send it in any version we've seen.

**Ask:** include `x-opencode-session: <stable-id-per-conversation>` on all outbound inference HTTP requests. Any stable UUID per conversation works. Happy to confirm from our telemetry once a build ships it. Thanks!

Moviw:
### Issue / context (if applicable)

Fixes #2317. OpenCode Go (the API behind the `opencode` provider) starts rejecting requests with no per-conversation ID from 09/05.

### Type of change

Bug fix

### What does this PR do?

Adds `x-opencode-session` alongside the existing `x-session-affinity` header in `session/llm.ts`, carrying the same stable `sessionID`, under the same non-ephemeral condition. Same idea as the existing header, just a second name for OpenCode Go's routing.

Added two tests to `llm-system-prompt.test.ts` (reusing its existing scripted-upstream mock): one asserts both headers carry the sessionID on a normal call, the other asserts neither header is sent on an ephemeral call. Ran `bun typecheck`, `bun lint`, and `bun test --cwd packages/opencode test/session/llm-system-prompt.test.ts` (8 pass, 0 fail). Reverting the header line locally makes the new test fail with the expected assertion, confirming it actually catches the regression.

### Checklist

### DIFF
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -812,7 +812,9 @@ const live: Layer.Layer<
         maxOutputTokens: params.maxOutputTokens,
         abortSignal: input.abort,
         headers: {
-          ...(!input.ephemeral ? { "x-session-affinity": input.sessionID } : {}),
+          // Stable per-conversation ID for upstream routing/optimization —
+          // OpenCode Go requires this from 09/05 (github.com/XiaomiMiMo/MiMo-Code/issues/2317).
+          ...(!input.ephemeral ? { "x-session-affinity": input.sessionID, "x-opencode-session": input.sessionID } : {}),
           ...(!input.ephemeral && input.parentSessionID ? { "x-parent-session-id": input.parentSessionID } : {}),
           ...input.model.headers,
           ...headers,
--- a/packages/opencode/test/session/llm-system-prompt.test.ts
+++ b/packages/opencode/test/session/llm-system-prompt.test.ts
@@ -558,3 +558,121 @@ describe("session.llm system prompt — memory-instructions guard", () => {
     }
   })
 })
+
+
+// https://github.com/XiaomiMiMo/MiMo-Code/issues/2317 — OpenCode Go rejects
+// requests with no per-conversation ID from 09/05. This reuses the module's
+// scripted-upstream harness (see the memory-instructions guard above) purely
+// to inspect the outbound request headers, not the system prompt.
+describe("session.llm outbound headers", () => {
+  test("a non-ephemeral stream call carries x-opencode-session and x-session-affinity", async () => {
+    const server = queueState.server!
+    const providerID = "alibaba"
+    const modelID = "qwen-plus"
+    const fixture = await loadFixture(providerID, modelID)
+    const request = waitRequest(
+      "/chat/completions",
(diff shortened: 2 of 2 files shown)


jiaxuan1101:
**Root cause: this is a regression inherited from upstream — and upstream is broken the same way**

MiMoCode 0.1.14 already sends two of the three correlation headers from this call site, but not the third. Verified against the released binary:

```
$ grep -a -c 'x-session-affinity'  "$(command -v mimo)"   # 1
$ grep -a -c 'x-parent-session-id' "$(command -v mimo)"   # 1
$ grep -a -c 'x-opencode-session'  "$(command -v mimo)"   # 0
```

Upstream history (`anomalyco/opencode`):

- **#36842** (merged 2026-07-14) added `x-opencode-project` / `x-opencode-session` / `x-opencode-client` in `packages/core/src/session/runner/llm.ts`.
- **#43188** (merged 2026-08-18, "restore session request headers") rewrote that same `http.headers` block to `x-session-affinity` / `X-Session-Id` / `x-parent-session-id`, dropping the three from #36842. Upstream `main` still carries only the second set today (lines 205–213).
- **#44752** (merged 2026-08-24) fixed the header set on `packages/opencode/src/session/llm/request.ts`, but the runner builds its own headers, so it stayed broken.

Consequence: upstream opencode itself started failing on 09/05 — issues #47438 and #47763 are both still open, and upstream's own one-line fix **#47473** (`+4/-0`, opened 2026-09-05) has 0 reviews.

So MiMoCode inherited this rather than introducing it, and #2327 is the local counterpart of #47473 — a regression fix, not new behavior. One delta worth noting: upstream's original #36842 sent three headers; #2327 (like #47473) restores only `x-opencode-session`. That may be all that's needed now, but #36842 is the reference if you want parity with where upstream was before the regression.

For scale: per #2317, 280 user orgs reach Go through MiMoCode, and Go has hard-rejected without this header since 09/05.

@yanyihan-xiaomi — the blocker looks like the first-time-contributor CI approval gate rather than the change itself (3 lines + tests, mergeable). A maintainer approval would let CI run.

---

**中文摘要**

这不是新功能，是从上游继承的回归。MiMoCode 0.1.14 在同一个调用点已发出 `x-session-affinity` 和 `x-parent-session-id`，只缺 `x-opencode-session`——上游在 08-18 用 #43188 的 header 集合**重写**了该调用点，覆盖掉 07-14 #36842 加的三个 header，导致上游 opencode 自己也在 09-05 起被 Go 硬拦（issue #47438 / #47763 均未关闭），其对应修复 PR #47473 同样是 4 行改动、至今 0 review。因此本 PR 应按**回归修复**处理。

heimoshuiyu:
opencode inference gateway now read `x-session-affinity`, could you check again?

PR #2334, opened 2026-09-04
feat(plugin): optional tool jsonSchema for bundled file tools

yanyihan-xiaomi:
## Summary

Add an optional `jsonSchema` field on plugin/file tools so bundled tools can ship a pre-derived provider-facing JSON Schema.

Self-contained file tools (esbuild-inlined zod) cannot rely on `.meta()` / `.describe()`: those annotations live in the tool copy's registry and are invisible when the engine derives the wire schema via `z.toJSONSchema(z.object(def.args))`. A nested `discriminatedUnion` then emits only `anyOf` without the sibling `type: "object"`, and some models stringify the whole tool-call envelope.

This change is **backward compatible**:

- `tool({ jsonSchema? })` in `@mimo-ai/plugin` — optional
- `Tool.Def.jsonSchema?` — optional
- `fromPlugin` forwards `jsonSchema` when present
- `prompt` / `llm-request-prefix` / experimental tool list prefer `item.jsonSchema` over re-deriving from `parameters`
- Tools that omit `jsonSchema` keep the previous behavior

## Test plan

- [x] `bun typecheck` (repo turbo typecheck on push)
- [x] `bun test test/tool/registry.test.ts` — new case: file tool `jsonSchema` is surfaced on `Tool.Def`
- [ ] Follow-up consumer (MiMo Desktop `tts_speech`) will export a wire schema with `voice: { type: "object", anyOf: [...] }`

## Notes for reviewers

- Shell-mode invocation still uses the shell-wrapped parameters and does **not** advertise a JSON-style `jsonSchema` (the pre-derived schema is authored against the JSON shape).
- No change to execute validation: `parameters` remains a zod object built from `args`.

### DIFF
--- a/packages/opencode/src/server/routes/instance/experimental.ts
+++ b/packages/opencode/src/server/routes/instance/experimental.ts
@@ -245,7 +245,7 @@ export const ExperimentalRoutes = lazy(() =>
           tools.map((t) => ({
             id: t.id,
             description: t.description,
-            parameters: z.toJSONSchema(t.parameters),
+            parameters: t.jsonSchema ?? z.toJSONSchema(t.parameters),
           })),
         )
       },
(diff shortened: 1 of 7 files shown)


yanyihan-xiaomi:
Closing: abandoning the plugin jsonSchema contract. Desktop will handle model anyOf/stringify at execute instead.

PR #2347, opened 2026-09-07
fix(actor): deliver terminal outcomes after execution settles

wqymi:
## Summary

Fix actor terminal delivery so parents receive the actual outcome without waiting on an already-ended child or confusing a running executor with a failed registry row.

- Remove initialization-time orphan sweeping based solely on another instance ID. Inactivity/timeout is not promoted to execution failure.
- Serialize the full in-process execution cycle across spawn, hooks, inbox continuations, notification, and cleanup. Reserve before exposing actor identity; publish the forked fiber without an intervening scheduler turn. Cancellation before fiber attachment remains effective.
- Queue inbox wakes until the full cycle ends, defer inbox consumption by hook re-entry, and drain each queued message once. Release continuation ownership only after its prompt runner and terminal delivery have finished.
- Preserve the main/gate result across postStop housekeeping. Non-cancellation gate/postStop LLM errors become warnings; a gate that could not verify completion reports partial/blocked according to task state, not unqualified task success. Genuine main failures remain failures and retain any available partial result.
- Persist the authoritative delivery in optional assistant.actorResult metadata. A nullable actor_registry.result_message_id column explicitly references this cycle's delivery; persist it before publishing terminal state. Failure wait reads only that reference, so absent partial results cannot fall back to a previous cycle. The additive migration creates no new table and does not duplicate result bodies.
- Preserve self-reported failed/blocked when gate verification fails, without weakening existing task-state precedence. Render multiline warnings in the existing TUI structured notification card, with theme colors and English/Chinese labels.
- Reuse one terminal notifier for spawn, continuation, and pending cancellation, retaining explicit system-agent/toast policies. Include lastOutcome in status, re-read terminal state before a wait timeout, and diagnose notification delivery errors.

This PR changes the MiMoCode engine, its TUI, tests, an additive nullable-column migration, and generated SDK type. It does not change Desktop implementation/UI code or engine-pin.json, add OS/process locks or leases, repair historical records, or automatically rerun failed work.

## Review follow-up

Commit `f91660054c` addresses the supplied review:

- **Result loss:** hook failures no longer discard the main result; warning and partial-result delivery are covered by tests, including persisted wait results.
- **Spawn/woken overlap:** the original race was reproduced with postStop held open while a woken turn wrote idle/success. Regression coverage now holds postStop, queues multiple messages, runs a hook re-entry, and sends another message at a continuation terminal boundary; no early terminal state, result mixing, duplicate turn, or lost wake is allowed.
- **Registration/cancellation:** reserve before onActorID and publish the fiber atomically; an onActorID cancellation regression verifies cancelled outcome and zero model requests.
- **Orphan semantics:** remove stale sweep comments. Do not adopt inactivity-derived failure; real crash recovery remains outside this PR.
- **Notification duplication/style:** share the terminal send path and format the touched test blocks. The main-session onInterrupt argument remains intentionally supported.

### Re-review follow-up

Commit `80aad242ec` addresses all three re-review items:

- Gate-failure fallback changes only success/unreported to partial; task-state blocked still takes precedence. Five integration cases change task state while the gate request fails and verify failed/blocked preservation.
- Both success and failure deliveries are persisted before terminal publication and referenced explicitly. A pending-wait integration verifies immediate partial delivery; regressions cover structured partials and previous success/failure receipts without current output. There is no failure-side scan for an older non-error assistant.
- Warning metadata is parsed outside Result/Partial result bodies and rendered by the real TUI component. Tests cover multiline warnings, metadata-looking body text, 32/80-column rendering, warning color, no-warning layout, and Chinese labels.

## Verification

From `packages/opencode`, run each selected test file in its own process:

```sh
for file in test/actor/*.test.ts test/inbox/*.test.ts \
  test/tool/actor.test.ts test/tool/actor-wait.test.ts test/tool/actor-cancel.test.ts \
  test/plugin/actor-hooks.test.ts test/task/gate.test.ts \
  test/session/checkpoint-splitover-integration.test.ts test/session/run-state-tuple-key.test.ts \
  test/effect/runner.test.ts test/effect/runner-warn-log.test.ts \
  test/cli/tui/actor-notification-warnings.test.tsx; do
  bun test --timeout 120000 "$file" || exit 1
done
bun typecheck
```

- **PASS: 323 tests across all 41 files, 0 failures**, each file run in a fresh process. No tests were skipped to get this result.
- PASS: package typecheck, SDK generation/build, and git diff --check.
- PASS: targeted oxlint, no errors; existing warnings remain.
- PASS: final Node engine build plus fresh **domestic** Desktop build; `react-subagent-detail.spec.ts`: **8 passed**. These UI regressions use controlled events/HTTP responses, not live-model end-to-end failure injection.
- Earlier single-process runs: the 305-test set passed; after adding the explicit first-spawn provider-error test, broad runs encountered auto-start/checkpoint timeouts, including one run exceeding the overall timeout. Both affected tests passed separately, and all 306 passed in the per-file run. The incomplete single-process runs are not claimed as green.
- RED evidence: original terminal regressions failed against the pre-fix implementation; the postStop result-preservation regression also failed against the initial PR commit. The overlap probe demonstrated idle/success while postStop was still blocked. Re-review regressions reproduced three failures on the previous implementation: failed/blocked overwritten by partial, and missing partial output in an already-pending wait. Warning parsing was also observed red before its fix.
- The latest Node build includes 43 migrations. Migration/build/UI validation used dedicated test databases; the running shared-data instance was not restarted or migrated for this follow-up.

## Earlier live validation and remaining limits

Before the review follow-up, a separate domestic Desktop development instance used shared existing data and **MiMo-X-Flash-Preview** through the real Desktop runHarness path. Parent and child records confirmed `xiaomi/mimo-x-flash-preview`; first completion, inbox continuation, running cancellation, and idempotent cancellation produced the expected notifications/wait outcomes. Children called no tools or modified business files. These observations are not presented as a new live-model rerun of this follow-up.

The empty-output probe returned a non-empty success and did not exercise real failure. Failure branches are covered by automated provider-error/hook-error injection. Shared data also contained a stale main-actor orphaned marker; its historical writer was not established. Existing records, older concurrently running engines, full process-crash recovery, and live cross-restart rendering remain outside the verified claim.

The primary agent performed the follow-up and self-review. No development/review subagents were used, per the user's instruction.

## TUI verification follow-up (2026-09-11)

Live TUI verification against isolated warning/failure cases (`script/actor-notification-cases.ts`), plus three polish fixes found while running it.

**Card polish**
- `fix(tui): localize actor notification status labels` — i18n keys replace the bare "sub-session completed/failed" badge (`子代理已完成` / `Subagent completed`).
- `fix(tui): compact actor notification card, keep full stack for main agent` — `parseActorNotification` now collapses summary/warnings to the first non-stack line (160-char cap). Stack frames stay on the synthetic message text for the main agent; the terminal card no longer renders a wall of `at foo (path:line)`.

**Startup robustness**
- `fix(provider): do not block TUI startup on models.dev` — `Data()`/`get()` read only local sources (pinned path, cache, build snapshot) and never await a network fetch. Background `refresh()` stays, but Flock timeout drops from 5min to 3s and a pinned `MIMOCODE_MODELS_PATH` skips network entirely. A hung models.dev request can no longer park first paint behind a 5-minute lock.

**Test fixtures**
- `test(actor): add process-lifecycle fixtures and notification cases script`.

**TUI screenshots (isolated cases)**
- warning: `✓ 子代理已完成 custom — PRESERVED-WARNING-RESULT` + one-line `警告: postStop: Error: Actor assistant failed: APIError` (no stack).
- failure: `✗ 子代理失败 custom — Error: Actor assistant failed: APIError` (no stack).

Verification: `parse-actor-notification.test.ts` + `actor-notification-warnings.test.tsx` = 31 pass; package typecheck pass; live TUI boot with `--preload @opentui/solid/preload`.

## Review findings follow-up (2026-09-12)

Three commits address the four review findings.

**1. Crash zombies (HIGH)** — `fix(actor): settle crashed zombies by abandon threshold on init`  
On registry init, running/pending rows whose last activity (or created time) is older than `DEFAULT_LIVENESS_ABANDON_MS` are settled as failure with a restart note. Time-based, not instance-id-based, so a live peer from another process is never falsely settled. Covers the single-process crash-recovery gap that orphan-sweep removal left open.

**2. Continuation clears result_message_id (MEDIUM)** — `fix(actor): persist continuation deliveries and skip cancelled drain`  
Inbox-woken `runTurn` now receives a settle that persists `actorResult` on the final assistant message and returns its id, matching spawn. Success continuation rewrites `result_message_id` after the running transition cleared it. Failure settle uses `lastFinal` so a turn that died with partial text still publishes that delivery. Pure API errors with no output still leave the column null (TP-R14-11: a failure without output must not reuse a previous success). Failed continuation notifications now carry partial `result`/`reportedStatus` when present.

**3. inboxWake drain before cancelled check (MEDIUM→LOW)** — same commit  
`execution.cancelled` is checked before drain. A cancelled wake no longer consumes messages for a turn that will not run; it still falls through to `runTurn` so `onExit` sends the cancelled notification. `executions.acquire` already serializes wakes, so concurrent drain races were not the issue.

**4. Cold start get() returns {} (LOW)** — `fix(provider): one-shot fetch on cold start with no cache or snapshot`  
Local-first remains the default (pinned path → cache → snapshot). If all three are missing, do one fetch bounded by the existing 5s AbortSignal so a fresh install still sees models; result is written to the cache. Pinned `MIMOCODE_MODELS_PATH` and `MIMOCODE_DISABLE_MODELS_FETCH` still skip network.

Tests: `registry` zombie sweep + result_message_id cases, `cancel-notification` continuation resultMessageID assertions, `models-nonblocking` cold-start fetch, full actor suite 237 pass / 0 fail, package typecheck pass.

### Residual drain race (2026-09-12)

A re-review of HEAD `eff3b0404` restated the four findings against the pre-fix commit. They are already closed on current HEAD (`92b79994e0` and earlier in this section). One residual from #3 remained: cancel can land after the pre-drain `cancelled` check and before the empty-drain short-circuit. `Actor.cancel`'s execution path does not notify on its own — only `runTurn.onExit` does — so the wake could return without a cancelled notification.

`4c724136c9` re-checks `execution.cancelled` after `drain === 0` and falls through to `runTurn` instead of short-circuiting. Tests: cancel-notification + registry + waiter = 52 pass / 0 fail; package typecheck pass.

### Re-review follow-up (2026-09-12, HEAD 4c724136c)

A second re-review restated three residuals after the abandon-threshold sweep landed.

**1. Same-process rebuild could settle live long-running actors (HIGH)** — closed in `6375e1f81d`.  
The time-based sweep ran on every layer init with no instance filter. A same-process rebuild (memo'd runtimes, overlapping tests) re-ran init while a fiber could still be alive; a running/pending row whose last part write was older than `DEFAULT_LIVENESS_ABANDON_MS` (long LLM step, no part writes) would be marked failure while the fiber still ran. The WHERE clause now requires `instance_id != PROCESS_INSTANCE_ID`, so only other-process abandoned rows are settled. Regression: `same-process rebuild does not settle a stale-but-alive running row`.

**2. Non-empty drain then cancel can drop wake content (MEDIUM, residual)** — accepted as follow-up.  
`inbox.drain` deletes Inbox rows and writes synthetic user messages before `continued` observes `cancelled`. If cancel lands in that window the turn does not run; the synthetic user text remains in the transcript for a later turn, but this wake is not executed. Closing it needs a requeue-or-don't-consume design beyond this PR. Empty-drain and pre-drain cancelled paths already notify via `runTurn.onExit`.

**3. models cold-start comment vs 5s fetch (LOW)** — comment corrected in `6375e1f81d`.  
Production ships `models-snapshot.js`, so TUI first paint does not wait on models.dev. A source checkout with neither cache nor snapshot may block up to 5s on the one-shot fetch; that trade-off is now documented in the comment. Outer `try/catch` replaced with `.catch`.

### Drain-cancel boundary closed (2026-09-12)

The last residual from the second re-review is closed in `a78c318e77`. Previously `inbox.drain` wrote the synthetic user message and deleted Inbox rows before the continuation observed `cancelled`, so a cancel in that window consumed the wake without running the turn.

`drain` now takes an optional `isCancelled` predicate and aborts **before any mutation** when it returns true — rows stay durable and the next non-cancelled wake still consumes them. The inbox-woken path in `prompt.ts` passes `() => execution.cancelled`. Unit coverage: cancelled drain leaves rows and writes no synthetic part; a later drain still returns 1; non-cancelled drain commits normally.

No remaining follow-ups on PR #2347. inbox + actor suites: **243 pass / 0 fail**.

### Mid-commit drain cancel closed (2026-09-12)

`2d90cbdafb` closes the last narrow window from the drain-cancel fix. `isCancelled` is re-checked after the synthetic user message is written and before each part / the inbox DELETE. If cancel lands there, `removeMessage` rolls the synthetic message back and Inbox rows stay durable — they remain the source of truth for the next non-cancelled wake.

No remaining follow-ups on PR #2347. Cross-instance abandon-threshold settling of other-process silent rows remains an intentional, documented trade-off.


wqymi:
## CI stabilization — f2aec754a1

Both previously failing shards now pass on the rebased branch.

Changes are limited to two test files:
- `execution-integration.test.ts`: each test owns the real `AppLayer` scope instead of reusing the process-wide `AppRuntime`. Async gate/wake/cancel callbacks use that test's captured service context, not a new or previously interrupted runtime.
- `run-state-tuple-key.test.ts`: replace the 10ms startup guess with Deferred signals from both running workers. The main runner remains held until the subrunner cancellation is verified, strengthening the isolation assertion without extending the timeout.

No production code, CI configuration, or test timeout was changed, and no new test skips were introduced.

Local verification: target 13 tests passed; same-process mixed run with prompt/MCP/server tests passed 110 tests with 3 pre-existing skips; typecheck passed.

Remote verification for `f2aec754a1d3547acda081dfa7a61a128a31bff1`: lint, typecheck, and all four unit shards passed.

https://github.com/XiaomiMiMo/MiMo-Code/actions/runs/34164837449

PR #2375, opened 2026-09-11
fix(models): keep local catalog available during best-effort refresh

wqymi:
## Summary
- Local models.dev catalog available before any network fetch (build-time snapshot + last-good cache).
- Host API: `ModelsDev.get/subscribe/startRefresh/refresh` on `node.ts`.
- Best-effort background refresh never blocks startup; invalid/failed responses keep last-good.
- Explicit operator `MODELS_PATH` is exclusive; never overwritten by network refresh.
- TUI: successful catalog publish debounces into `Instance.disposeAll()` (skips in-flight) so the existing `server.instance.disposed` → bootstrap path refreshes the provider list.

## Review follow-ups (cumulative)
- Entity set only from authoritative cache/fetch; snapshot fills fields, never revives dropped entities
- Invalid `MIMOCODE_MODELS_PATH` fails closed (`{}` + `onError`); no public fallback
- `refresh(true)` not swallowed by in-flight non-force; owed force still runs after a failed flight
- `MIMOCODE_DISABLE_MODELS_FETCH` gates auto/`startRefresh` only
- Exclusive explicit PATH short-circuits `update()` (no HTTP / no shared-cache rewrite)
- `reset()`: generation guard on both `load()` and `publish()` settles
- Embedders that skip `src/index.ts` / TUI worker must call `startRefresh()` after wiring fetch

## Testing
- `bun test test/provider/` — PASS
- `bun typecheck` in `packages/opencode` — PASS
- Covers: snapshot-only deletion, force-join, force-after-error, disabled×force, explicit fail-closed/no-fetch, reset re-read, catalog-reload debounce

## Notes
- Based on current `origin/main`. Rebase conflict in `models.ts` (#2336 removed voice fields) resolved by keeping extracted `models-catalog.ts`/`models-schema.ts`; optional `voice_*` remain for operator speech metadata (noted as overlay debt).
- `get()` structuredClone is intentional (callers mutate); freeze is a later opt.


wqymi:
Rebased onto current `origin/main` (`f11e35ede4`). Conflict in `models.ts` resolved by keeping the new `models-catalog.ts`/`models-schema.ts` structure; optional `voice_*` fields retained for Desktop speech metadata (still marked as overlay debt).

Verification after rebase:
- `bun test test/provider/` — 549 pass / 0 fail
- `bun typecheck` in `packages/opencode` — pass

HEAD: `4976642e44`

wqymi:
Added TUI catalog hot-reload (`8b0cd74478`):

- `ModelsDev.subscribe` → debounced `Instance.disposeAll()` (skips in-flight)
- reuses connect path: `server.instance.disposed` → TUI `bootstrap()`
- unit: `models-catalog-reload.test.ts`

Desktop pin in !4267 updated to same commit.

PR #2406, opened 2026-09-16
fix(retry): stop request-phase ladders from stacking session.status reconnect frames

wqymi:
## Summary

When the model upstream is unreachable (proxy/DNS failure, invalid baseURL), `session.status{type:"retry"}` could fire far denser than exponential backoff suggests.

Instrumentation with invalid provider `baseURL` (`http://127.0.0.1:1/v1`) reproduced **20** `session.status{retry}` publishes in ~32s **without maxMode**:

- nested `request×4 + stream` per processor cycle
- request ladders reset to ~200ms every outer retry
- `Cannot connect to API` rode a ~2s-class stream/server ladder, packing ~4 outer cycles → 20 status frames
- a pure network-kind nested ceiling (5s→60s) would be lower (~15); this error class packs denser

Consumers that key off `session.status{type:"retry"}` (e.g. TUI reconnect chrome) therefore saw a streak that did not look like exponential backoff.

### Fix

1. **`session.status{retry}` is processor-owned** — `llm.ts` request-phase no longer calls `status.setRetry`; it only publishes `Session.Event.RetryAttempt` (phase-local diagnostics). User-visible retry status is the processor stream ladder.
2. **Max-mode does not set `ephemeral:true` for status control** — `ephemeral` also skips plugins, session-affinity headers, OTel `functionId`, and system assembly. Ownership alone stops request-phase status stacking.
3. **Density telemetry** — `session.status` retry publishes log `sessionID/attempt/phase/waitMs`.
4. **Docs** — `retry-coordinator.md`: processor owns status; request `RetryAttempt.attempt` is phase-local; quieter status during request ladders is intentional.
5. **Tests** — invalid-baseURL density integration; max-mode source guard against `ephemeral: true`; arithmetic notes; processor-effect status ownership.

## Verification

- `bun typecheck` (packages/opencode) — pass
- focused `bun test` retry / max-mode / processor-effect publish-status — pass
- `retry-density-invalid-baseurl.test.ts` — post-fix: stream-only status frames, growing waits (~4 frames / 32s)


wqymi:
## Author note — CR round 2

Independent re-review **CR-PASS** on HEAD `0554a589ff`.

Round-1 blockers verified closed:
- no max-mode `ephemeral` overload
- no dead `SessionStatus` in `llm.ts`
- arch doc attempt semantics aligned (processor global vs request phase-local)
- field “20 frames” narrative reconciled with ~2s server/stream ladder packing

Post-fix density (reviewer re-run): ~33s, `session.status` frames = **4**, all `stream`, waits 2.1s→16.7s.

### 🟡 non-blocking (accepted as follow-ups, not merge gates)

1. max-mode guard is a source-string scan — replace with runtime mock later if it flakes on refactor
2. parallel ensemble still emits diagnostic `RetryAttempt` on request failures (UI does not consume this)
3. `status.ts` retry slog.info may be noisy under storms — keep for field density forensics for now

Merge remains for a human confirm after review; this PR is not auto-merged.

wqymi:
## Author — CR round-2 🟡 closed (policy: non-blocking must-fix)

User policy: even non-blocking review items are fixed before close. Round-2 🟡 → HEAD:

1. **Runtime max-mode contract test** (replaces source-string scan)  
   `max-mode-retry-status-ephemeral.test.ts` captures real `llm.stream` inputs for candidate/judge and asserts `quietRetryDiagnostics === true` and `ephemeral` falsy.

2. **Ensemble request RetryAttempt noise**  
   New narrow `StreamInput.quietRetryDiagnostics` — max-mode candidate/judge set it; request-phase skips `Session.Event.RetryAttempt` without overloading `ephemeral` (plugins/affinity/OTel/system untouched). Processor-owned `session.status` unchanged.

3. **status.ts slog flood**  
   `session.status retry publish` telemetry demoted `slog.info` → `slog.debug` (fields kept for field forensics).

Arch doc updated for `quietRetryDiagnostics`.  
Verification: `bun typecheck` + max-mode/econnreset/density arithmetic tests green.

Not an auto-merge; human confirm still required.

PR #2407, opened 2026-09-16
fix(session): persistent exponential retry for recoverable errors

wqymi:
## Summary
- `network` / `rate_limit` / `server` are recoverable: persistent exponential backoff regardless of request vs stream phase
- `budgetFor` order: max-mode scopes → recoverable kinds → phase=request (request budget no longer short-kills DNS/429/5xx)
- Network fingerprints: `ENOTFOUND`, `ENETRESET`, `getaddrinfo`, `Cannot connect to API`
- Auth / quota / abort / tool side-effect boundaries remain terminal

## Why
Field sessions wrote terminal `assistant.error` after ENOTFOUND / rate-limit / 5xx because request-phase budget (≤4 / 30s) shadowed the persistent network policy. Product contract: longer wait improves recovery odds; do not kill recoverable kinds in a short window.

## Verification
- `cd packages/opencode && bun test test/session/retry.test.ts test/provider/error.test.ts` → pass
- `bun test test/session/llm-retry.test.ts test/session/run-state-tuple-key.test.ts test/provider/transport-shape-covers-legacy-network-codes.test.ts` → pass
- Shape classification locks coverage of every legacy `RETRYABLE_NETWORK_CODES` / `RETRYABLE_NETWORK_MESSAGES` entry
- `connection timed out after 30s` is recoverable network (message-only timeout without connection/network language stays non-retryable)
- `bun typecheck` → pass

## Notes
- max-candidate / max-judge keep their own budgets
- Desktop docs MR: mimo-innovation/mimo-desktop!4336


wqymi:
Update: transport classification is now shape-based (UND_ERR_* / E* outside local-fs deny-list / message family) instead of a closed retryable-code allow-list. Tests: 92 pass.

PR #2421, opened 2026-09-17
fix(provider): tolerate missing tool call IDs in openai-compatible streams

MiMoHardFather:
## Summary

Some OpenAI-compatible endpoints stream a tool call's `function.name` and `arguments` before its `id`, or never send an `id` at all. The upstream `@ai-sdk/openai-compatible@2.0.41` parser threw `Expected 'id' to be a string.` on the first delta, aborting the whole tool call.

This updates the repo patch for `@ai-sdk/openai-compatible@2.0.41` (both `dist/index.js` and `dist/index.mjs`):

- Buffer `id`, `name`, `arguments` and `thought_signature` per index until both `id` and `name` are known before emitting `tool-input-start`, so a late upstream `id` stays authoritative.
- On flush, fall back to `generateId()` (matching `doGenerate`) and reuse that single id across `tool-input-start`, `tool-input-delta`, `tool-input-end` and `tool-call`.
- Missing `function.name` still throws; a non-string `id` is still rejected by the chunk schema.
- The existing `lastValidDelta` fallback is preserved.

Only providers routed through `@ai-sdk/openai-compatible` are affected. The in-repo Copilot copy and other `@ai-sdk/*` packages keep upstream behaviour.

## Tests

- New `packages/opencode/test/provider/openai-compatible-tool-id.test.ts` runs against both ESM and CommonJS builds: undefined/null/empty ids, late ids, empty opener, interleaved indexes, providers that omit `index`, thought signatures, missing name still rejected, non-string id still rejected, and an end-to-end `streamText` round trip that checks the assistant `tool_calls[].id` matches the tool message `tool_call_id`.
- `bun test test/provider test/mcp/openai-compatible-args-patch.test.ts test/llm-server/streaming.test.ts` in `packages/opencode`: 610 pass, 0 fail.
- `bun typecheck` in `packages/opencode`: pass.
- Patch verified to apply cleanly to pristine 2.0.41 and to match the installed `node_modules` byte for byte. `bun.lock` unchanged.


patches/@ai-sdk%2Fopenai-compatible@2.0.41.patch
```diff
@@ -101,12 +168,40 @@ index dca128d..abe027f 100644
...
+               (toolCall2) => !toolCall2.hasFinished
+             )) {
++              if (!toolCall.function.name) {
++                throw new import_provider3.InvalidResponseDataError({
++                  data: toolCall,
++                  message: `Expected 'function.name' to be a string.`
++                });
++              }
++              if (!toolCall.hasStarted) {
++                // Match doGenerate's fallback, using one ID for every tool event.
```
MiMoHardFather: **[suggestion]** The flush comment ("Match doGenerate's fallback, using one ID for every tool event.") leaks design rationale into shipped dist, and it is not quite true: `doGenerate` uses `toolCall.id ?? generateId()` (keeps `""`), while this path uses `toolCall.id ||= generateId()` (treats `""` as missing, which the tests want). Same comment is duplicated on the ESM side.

**Suggestion:** Drop the doGenerate cross-reference. If a comment is needed, state the constraint only: assign the generated id onto `toolCall` before enqueueing so start/delta/end/call share it. Apply the same edit in the ESM hunk.

PR #2460, opened 2026-09-21
fix(mcp): bound tool metadata before persistence

MiMoHardFather:
### Type of change

Bug fix

### What does this PR do?

Large MCP results were truncated in `output`, but their full `structuredContent`, `_meta`, and legacy metadata still reached the session database, including nested `exec` results. A roughly 1 MB structured response could therefore leave a roughly 1 MB database record despite a tiny text preview.

Apply a shared 50 KiB inline metadata budget, with oversized fields archived as JSON in tool-output storage and replaced by file references. Apply the same handling to progress updates and error results. Keep private references under `_meta` and carry the full structured value to `exec` through a symbol-keyed runtime field that JSON serialization omits. Existing database rows are not migrated.

After rebasing onto the latest `main`, 157 related tests passed with no failures. Package type checking and the pre-push workspace type checks also passed. Tests run from `packages/opencode`:

- `bun typecheck`
- `bun test test/session/prompt-effect.test.ts --test-name-pattern MCP --timeout 30000`
- MCP metadata, normalization, and progress tests; `exec` and sharing tests; tool-gate unit, cancellation, hook, and orchestration tests.

The new regressions cover large UTF-8 payloads, direct and nested calls, success and error results, progress events, archive contents, private metadata, and full structured access inside `exec`.

### Checklist


Water-Run:
MimoHardFather?

